Issue #97, opened 2012-04-09
`self.` is often not needed and therefore better left off.


marcandre:
Especially for developers coming from Javascript, I'd suggest stating that the `self.` in `self.my_method` is usually not needed and should be left off for brevity's sake.

I'll write something up if you'd like me to.

dsferreira:
My thoughts on this matter:

I believe public interface should be self contained.
To emphasise this I like to enforce self in public methods inside their classes or modules so that their usage is conscious enough.
This way the developer is forced to pay for the cost of an extra method exposed in the public interface.
Code readability is also improved by showing to the reader the nature of each one of the methods in use.

PR #104, opened 2012-05-05
Avoid self

marcandre:
See #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,6 +479,55 @@ You can generate a PDF or an HTML copy of this guide using
     end
     ```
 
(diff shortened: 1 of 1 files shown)


dsferreira:
My thoughts on this matter:

- I believe public interface should be self contained.
- To emphasise this I like to enforce self in public methods inside classes or modules so that their usage is conscious enough.
- This way the developer is forced to pay for the cost of an extra method exposed in the public interface.
- Code readability is also improved by showing to the reader the nature of each one of the methods in use.

Issue #169, opened 2013-05-12
multiline chaining


lcowell:
I have a question about the multiline chaining example. You say:

> When continuing a chained method invocation on another line keep the . on the second line.

``` ruby
# bad - need to consult first line to understand second line
one.two.three.
  four

# good - it's immediately clear what's going on the second line
one.two.three
  .four
```

However, the example labelled good does not work for me (using 1.9), while the example labelled bad does. I thought the dot at the end of the line was necessary in order to tell the ruby interpreter that the expression continued.

Here's an example:

``` ruby
1.to_i
  .to_s
```

> syntax error, unexpected '.', expecting $end .to_s

Update:
OK, I've just figured out the issue is a 1.9 vs. 2.0 difference in syntax. The good example is only valid ruby in 2.0. I think there should be a note added to avoid confusion for people using 1.9. What is the recommended syntax for 1.9 ?

Thanks guys.

PR #176, opened 2013-05-28
include '.' on first line of multiline chained method invocations

lcowell:
This changes where the `.` lives from the second line to the first line in the expression. Having no `.` at the end of the first line makes that line appear to be complete, but you must read ahead to the next line in order to see that the expression continues.

This closes issue #169

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,16 +309,16 @@ Translations of the guide are available in the following languages:
              - 2
     ```
 
-* When continuing a chained method invocation on another line keep the `.` on the second line.
+* When continuing a chained method invocation on another line, include the `.` on the first line to indicate that the expression continues.
 
     ```Ruby
-    # bad - need to consult first line to understand second line
-    one.two.three.
-      four
-
-    # good - it's immediately clear what's going on the second line
+    # bad - need to read ahead to the second line to know that the chain continues
     one.two.three
       .four
+
+    # good - it's immediately clear that the expression continues beyond the first line
+    one.two.three.
+      four
     ```
 
 * Align the parameters of a method call if they span more than one line.


dsferreira:
I believe this style is misguiding.
I also believe we should avoid long method chains.
And provided we have self contained method chains I would prefer a single line approach:
foo(arg1).bar(arg2).baz(arg3, arg4)
I would also add a maximum configuration on this upon which an error/alert should be raised by rubocop.
In my opinion the default maximum should be 3.

lleolin:
@kballenegger 's approach is arguably a nice compromise between the two styles. The trailing slash indicates that the statement continues, while still allowing one to use preceding dots, and it still works in IRB. Additionally, it doesn't require any configuration changes to Rubocop. I'm happy with it.

I agree that we should be avoiding long method chains in most scenarios. However, sometimes it's not the number of methods being chained that causes one to have to continue a statement on the next line, but merely the number of characters in a statement.

roryokane:
I have compiled all the arguments I could find in this thread into a comprehensive list of pros and cons. All four styles described in this thread are compared:

* `.` at line end
* `.` at line beginning
* `.` at line beginning, indented to the previous `.`
* `\` at line end, `.` at next line’s beginning

The pros and cons are kind of long, so I’ll just [link to them on Stack Overflow](http://stackoverflow.com/a/29482762/578288).

justin808:
@bbatsov is there a rubocop setting for this?

bbatsov:
> s the default to be beginning of line or end of line? Seems to be.

Yes. See `default.yml` for alternatives.

kballenegger:
Kenneth Ballenegger has read this message at 6:48 PM on 4/15/2015.Read
receipt sent by Gmelius <http://gmelius.com/?src=mail>

bb010g:
[**@focused**](https://github.com/focused) commented on [Jun 9, 2013, 2:21 AM PDT](https://github.com/bbatsov/ruby-style-guide/pull/176#issuecomment-19163168 "2013-06-09T09:21:50Z - Replied by Github Comment Enhancer"):
> When I start using an awesome gem for monads [monadic](https://github.com/pzol/monadic#either), I realized that sometimes it is better to write dot at the end of line. Look (`>=` is an alias for `bind`):
> 
> ```ruby
> Either(operation).
>   >= { successful_method }.
>   >= { failful_operation }</pre></div>
> ```

I've always seen it in Haskell as (looks better with `bind` instead of `>=` here IMHO)
```ruby
monadic_operation.bind ->(foo)
  {bar foo, 1}.bind ->(biz)
  {baz biz, 2}
```
which would be equivalent to the following in Haskell do-notation
```haskell
do
  foo <- monadicOperation
  biz <- bar foo 1
  baz biz 2
```

kashif-umair:
I have a question about blocks with these chaining methods. I'm using Rails so my example is specific to Rails but rails is Ruby after all. So here it is:

```ruby
user.posts.includes(:comments).where(published: true).where.not(post_category: 'Private').each do |post|
  # do something with post here
end
```

What is the best way to write the above code? I do something like this:

```ruby
user.posts.includes(:comments).
where(published: true).
where.not(post_category: 'Private').each do |post|
  # do something with post here
end
```

I prefer to use trailing dots but if look at my code, I have not indented line number 2 and 3 because if I indent them then I have to add one more indentation level to `each` block.

equivalent:
I would say:

```ruby
user
  .posts
  .includes(:comments)
  .where(published: true)
  .where.not(post_category: 'Private')
  .each do |post|
     # do something with post here
  end
```

or

```ruby
user.
  posts.
  includes(:comments).
  where(published: true).
  where.not(post_category: 'Private').
  each do |post|
     # do something with post here
  end
```

I would prefer the first version

reason for having it in separate lines is that it's easier to alter (e.g. you add `.order(:created_at)` the commit message will be just one line in git log instead of long line where is not clear what have changed.

franklinyu:
> I have not indented line number 2 and 3 because if I indent them then I have to add one more indentation level to `each` block.

@kashif-umair I think you may as well just indent line 2 and 3, making `each` block indented one more level, because that makes sense: one level for continuation, another for nesting. In ruby code, I believe indenting 5 levels are very common.

cyoce:
@bbatsov can you please put option C in the style guide? It is a nice compromise.

    one \
      .two

mhenrixon:
Very interesting discussion! Personally I always had a very strong preference for the first option with the leading `.` but for me it is just how my brain is wired. It is also how I prefer to work with other languages like SQL where I prefer the `, ` to be in the beginning of the line to more easily be able to add and remove lines without having to also worry about adding a trailing comma somewhere.

Never thought of the option `C` that is definitely my new preference! 

Many thanks to everyone for the lengthy and thorough discussion.

Startouf:
Ah wouldn't it be just so great if Ruby could read syntax like this one

```
some_data.
.map { |d| d**2 }.
.sort_by { |d| d.odd? }.
.reject { |d| d.zero? }.
.reduce(:+)
```
it would be a nice compromise (that would looks better than the one mixing `\`and `.`)

As far as I know the `..` operator is only used for date ranges but never broken by a `\n` so shouldn't that be theoretically feasible in future versions of Ruby ?

Also, and this most likely a personal feeling, but I don't like the indentation after the first line breaks, whatever the choice for the placement of the dot. But in case of leading dots it feels obvious enough that the chain continues from the previous lines that I don't see the pros of indentation. 

On the contrary I would argue extra indentation makes it harder to close blocks. Take the following example of chaining with multiline blocks (putting aside the code smell), it makes it harder to realize if you're missing an `end` or not. I've quite several times struggled to find the missing `end` of code blocks, and having code that offsets itself for no reason makes it more error-prone. Maybe my linters are just bad, but often I would just get a red rubocop error at the end of the file and I'd have to find the missing/extra `end` myself

```
def get_questions
  Question.
    where(:score.gte => 42).
    pluck(:answer).
    select do |answer|
      answer.is_the_answer_to_life_and_everything?
    end.uniq
  # If not paying attention I'd want to put an `end` here
end
```

PR #215, opened 2013-08-18
Use short-circuited return statement instead of next blocks

girijabrahme:


guigs:
https://github.com/bbatsov/ruby-style-guide/issues/602

PR #284, opened 2014-02-13
Fix contradiction in percent literals rule

abotalov:
First rule says "Use %()" and the second rule says "Avoid %q()"

### DIFF
--- a/README.md
+++ b/README.md
@@ -2499,7 +2499,7 @@ this rule only to arrays with two or more elements.
     %(<tr><td class="name">#{name}</td>)
     ```
 
-* Avoid `%q` unless you have a string with both `'` and `"` in
+* Avoid `%()` unless you have a string with both `'` and `"` in
   it. Regular string literals are more readable and should be
   preferred unless a lot of characters would have to be escaped in
   them.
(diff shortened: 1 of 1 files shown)


strand:
@lee-dohm Here are the percentages:

|   | Rubinius | Puppet | Sass | Rails |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| without a q | 84.6% | 34% | 2.7% | 94.4% |
| with a q       | 15.4% | 66% | 97.3% | 5.6% |

Would it be sufficient to revise that sentence to read: "Avoid `%()` or the equivlant `%q()` unless you have a string with both `'` and `"` in it."

onebree:
while you can argue that `%q()` is longer, it is by only 1 character. I think that using the q is more descriptive and telling of what the LOC does. Yes, if you know Ruby you should know that `%()` is the same, but it is good for novice coders and those new to a project.

Is there any noticeable difference between `%()` and `%q()`? If both are used heavily in the community, then why post a style guide for it? Other than simply picking one and sticking to your guns.

onebree:
revisiting this PR, I agree with @strand's last comment. Quotes should be preferred in any case, unless the string is complicated enough that its original form breaks the flow somehow. If an internal `'` closes the string by accident (or something), then use the literal.

In regards to losing the Q, I stick to my original opinion above.

go2null:
As a Ruby neophyte, the `%` took some googling to find.

It also is very similar to [Perl-style special variables](https://github.com/bbatsov/ruby-style-guide#no-cryptic-perlisms) which are recommended to be avoided in favor of *human-friendly aliases*.

bbatsov:
> Would it be sufficient to revise that sentence to read as follows? "Avoid %() or the equivlant %q() unless you have a string with both ' and " in it."

Guess that'd be fine.

strand:
I've made PR #615 with the suggested change.

strand:
Please close as PR #615 resolves this concern.

PR #310, opened 2014-04-07
Classes: use attributes directly instead of call accessors methods.

korun:
Hello!

I added a few rules for use of `attr` family methods inside a class definition.

In example `name` is equivalent of `self.name` and this is a complete method call.

`@name` is more readable, and faster in use.

You can simple test it something like this:

``` ruby
class TestClass
  attr_reader :name

  def initialize
    @name = 'test_name'
  end

  def test1
    t = Time.now
    100_000_000.times{ @name }
    Time.now - t
  end

  def test2
    t = Time.now
    100_000_000.times{ name }
    Time.now - t
  end
end

x = TestClass.new
puts x.test1
puts x.test2
```

Moreover you can never shade `@name` of a some local variable. :smirk:

**UPD**: see improved test [below](https://github.com/bbatsov/ruby-style-guide/pull/310#issuecomment-40653771).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2278,6 +2278,28 @@ condition](#safe-assignment-in-condition).
   attr_reader :one, :two, :three
   ```
 
+* <a name="direct-attrs"></a>
+  Use attributes directly instead of call accessors methods.
+<sup>[[link](#direct-attrs)]</sup>
+
+  ```Ruby
+  class Robot
+    attr_reader :name
+
+    # some methods omitted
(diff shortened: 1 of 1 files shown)


strand:
I prefer the barewords style as well, @bbatsov what do you think?

bbatsov:
I think this is all too subjective and it's probably best to avoid adding a rule on subject to the style guide.

PR #325, opened 2014-06-25
Using unless for all cases of 'if not' isn't natual to an English speaker

cheerfulstoic:
It is much better to use either if not or unless depending on if the condition is exceptional.

I feel like following this rule (while it make take some getting used to), leads to generally more readable code and more thoughtful coding

### DIFF
--- a/README.md
+++ b/README.md
@@ -788,21 +788,32 @@ Translations of the guide are available in the following languages:
   end
   ```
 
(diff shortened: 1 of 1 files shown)


strand:
Can we close this PR, as #329 covers a very similar issue?

cheerfulstoic:
Certainly

PR #350, opened 2014-08-13
Array() instead of conditional

dimus:
This approach does not work well with hashes. Array() modifies
hash making it a multi-dimentional array.


strand:
I think it would be clearer to phrase it like this:

> Note: this approach does not work if a variable is either a hash or an array of hashes, as Array() converts the hash to an array before wrapping it in an array, thus creating an array of arrays of key-value pairs.

bbatsov:
Closing due to lack of activity.

Issue #351, opened 2014-08-17
Contradiction in examples of Hash#fetch


abotalov:
From https://github.com/bbatsov/ruby-style-guide#hash-fetch-defaults:

```
# good - fetch work correctly with falsy values
batman.fetch(:is_evil, true)
```

And then just after it:
> Prefer the use of the block instead of the default value in Hash#fetch

PR #353, opened 2014-08-20
Hash#fetch with default value also has sometimes right to life

abotalov:
Fixes https://github.com/bbatsov/ruby-style-guide/issues/351

### DIFF
--- a/README.md
+++ b/README.md
@@ -2975,7 +2975,7 @@ resource cleanup when possible.
   ```
 
 * <a name="use-hash-blocks"></a>
-  Prefer the use of the block instead of the default value in `Hash#fetch`.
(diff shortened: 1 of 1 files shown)


bbatsov:
@abotalov ping

README.md
```diff
@@ -2848,7 +2848,7 @@ Translations of the guide are available in the following languages:
   ```
 
 * <a name="use-hash-blocks"></a>
-  Prefer the use of the block instead of the default value in `Hash#fetch`.
+  Prefer the use of the block instead of the default value in `Hash#fetch` if non-block evaluation may have side effects or be expensive.
```
abotalov: fixed

Issue #352, opened 2014-08-19
Multi-line comment spacing inconsistency


fishermand46:
The rule about multi-line comments says: "If multiple lines are required to describe the problem, subsequent lines should be indented two spaces after the #"

However, in the example:

``` ruby
# FIXME: This has crashed occasionally since v3.2.1. It may
#   be related to the BarBazUtil upgrade.
```

the second line is indented three spaces after the #.

fishermand46:
Any thoughts on this? Should it read "three spaces after the #" or should the example be intended two spaces?

LavirtheWhiolet:
:+1: to @fishermand46.

bbatsov:
Indeed. I guess we just made a typo with the example.

tejasbubane:
@bbatsov I think this has been fixed in https://github.com/rubocop-hq/ruby-style-guide/pull/346.

PR #357, opened 2014-09-02
Fix comment indent specification.

fishermand46:
The specification for multi-line comments said to indent two spaces after the '#', but the example shoed three spaces. I fixed the specification to be consistent with the example.

Fixes Issue #352.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1981,7 +1981,7 @@ Translations of the guide are available in the following languages:
 
 * <a name="indent-annotations"></a>
   If multiple lines are required to describe the problem, subsequent lines
-  should be indented two spaces after the `#`.
+  should be indented three spaces after the `#`.
 <sup>[[link](#indent-annotations)]</sup>
 
   ```Ruby


erran:
> subsequent lines should be indented <del>**two**</del><ins>**three**</ins> spaces after the `#`.

@fishermand46's change is correct in that it is three spaces after the `#`. Possibly adjust the comment to:
> subsequent lines should be indented two spaces after the **beginning of the annotation** (three spaces after the `#`).

bbatsov:
@erran Let's merge #346 as it was submitted first. I had totally forgot about it.

snowe2010:
Does this indentation apply to regular comments as well or only annotations? The guide is not clear about it and I didn't want to open a new issue for such a little thing.

onebree:
In Vim, when I begin a comment (no matter initial indentation) with `# `, hitting enter will begin the next line with the same 2 characters.

I think to be clear, any commenting rules should be in one section.

1. Begin _all_ comments with `# `
2. Except for in-line comments, try to write block comments out as sentences.
2. Avoid using `=begin`/`=end` as it is generally unreadable, and does not allow indentation
3. When comments need to include an example, indent one additional time.

```ruby
string = "Hello"    # sample inline comment

# This is a sample GOOD comment.

#A sample *okay* comment

=begin
  Sample bad comment blocking
    Especially with leading white space
=end

# sample block comment with indents
# Sample of regular route:
#   match 'products/:id' => 'catalog#view'
# Keep in mind you can assign values other than :controller and :action
```

PR #367, opened 2014-09-22
Semantic indent and alignment

IronSavior:
Use whitespace to communicate structural significance.


IronSavior:
We haven't heard any objections and it's been several months. Would you agree that it's safe to merge this one?

PR #368, opened 2014-09-22
Add acceptable use case for !!

IronSavior:
The original sentiment is correct for most contexts.  However, it is still acceptable to use `!!` when you don't necessarily want to expose the value of the expression, but you do want to expose whether or not it is nil/false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -800,7 +800,9 @@ Translations of the guide are available in the following languages:
   ```
 
 * <a name="no-bang-bang"></a>
-  Avoid the use of `!!`.
+  Avoid the use of `!!` for most contexts.  It is acceptable to use `!!` to
+  coerce a method's return value to truthiness.  This can be useful if you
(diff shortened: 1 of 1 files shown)


strand:
@bbatsov Is `!!` acceptable for coercing `nil` or `false` to `false`, and all other values to `true`?

antstorm:
#591

bbatsov:
> Perhaps it is a matter of taste, I submit this case as being merely an "acceptable" edge case rather than being canonically The Right Thing.

Probably. While I understand your reasoning that's very hard to enforce automatically and that's why I dislike it. Same thing with using `and` and `or` for flow of control.

RomanADavis:
They both seem kind of bad, would be tempted to monkey patch in:

```
class Object
  def to_bool
    self ? true : false
  end
end
```

The nature of monkey patching makes that non ideal. Has anyone asked the core ruby devs to add something like this?

PR #375, opened 2014-10-04
use private_class_method to make class methods private

akshaymohite:


Lucasosf:
I think using private_class_method isn't clear enough such as the current way.

freemanoid:
What about using `private_class_method` inline:
```ruby
class TestClass
  private_class_method def self.some_method
    # body omitted
  end
end
```

philoserf:
I do not see the benefit. I found this quote in my search for 'prior art' on the question, "Makes existing class methods private." It struck a cord with me.

I believe the style marked 'bad' in this PR is the common case. If I need included code, I do not own to 'become' private i would use `private_class_method` to re-scope it.

bbatsov:
> I believe the style marked 'bad' in this PR is the common case. If I need included code, I do not own to 'become' private i would use private_class_method to re-scope it.

The "common case" actually does nothing, but this is a logic error, not something that should be in a style guide.

onebree:
Throughout our company's app, we include `private`, followed by methods for params, page name, and a couple other things... According to what @philoserf said, this means stating `private` does nothing to the methods that follow.

If so, then why include `private`? What _does_ it do?

bbatsov:
`private` works on instance methods, but not on class methods.

jaredbeck:
> `private` works on instance methods, but not on class methods.

This is a common mistake. As such, I would like some way to help my projects avoid this mistake.  Maybe it could be implemented in rubocop?

> .. this is a logic error, not something that should be in a style guide.

If applied consistently, that makes sense, even though [other popular style guides][1] do chose to cover logic errors.

[1]: https://github.com/airbnb/javascript

bbatsov:
> This is a common mistake. As such, I would like some way to help my projects avoid this mistake. Maybe it could be implemented in rubocop?

RuboCop will warn you if you do something like this. :-)

jaredbeck:
>> This is a common mistake. As such, I would like some way to help my projects avoid this mistake. Maybe it could be implemented in rubocop?
> 
> RuboCop will warn you if you do something like this. :-)

Thanks! I tried it out, but it doesn't catch all permutations of this mistake (https://github.com/bbatsov/rubocop/issues/2505)

PR #392, opened 2014-12-03
Change inside of method (related #378)

marocchino:


sethkrasnianski:
Looking back at this, I believe it was incorrect to merge it. The correction was not accurate. Referring to my original PR

>When writing a method's signature (declaring the method), the values being passed to a method are called parameters, not arguments. When calling a method, the values passed to the call are formally known as arguments. This is a general misnomer in the programming community. I've corrected all relevant instances.

`to_str.#{unsafe_method}(*params, &block)`
and
`find_by(prop, *params, &block)`
are both calling respective methods, not declaring them.

marocchino:
> are both calling respective methods, not declaring them.
Yes, it is. But it occur `NameError`, when it called.

sethkrasnianski:
sure, that may happen, but it's still a value passed to the method... right?

PR #402, opened 2015-01-07
Use double intent when continuing a line

fishermand46:
When continuing a line onto a second, I find it helpful to use a double indent so as to distinguish it from a normal start of a block.


fishermand46:
Decided to start a general discussion with #402.

PR #404, opened 2015-01-18
Slightly improve no-extend-struct-new wording and add code example

mmozuras:
@bbatsov

### DIFF
--- a/README.md
+++ b/README.md
@@ -2301,10 +2301,19 @@ condition](#safe-assignment-in-condition).
   ````
 
 * <a name="no-extend-struct-new"></a>
-  Don't extend a `Struct.new` - it already is a new class. Extending it
-  introduces a superfluous class level and may also introduce weird errors if
-  the file is required multiple times.
+  Don't extend an instance initialized by `Struct.new`. Extending it introduces
+  a superfluous class level and may also introduce weird errors if the file is
+  required multiple times.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2301,11 +2301,21 @@ condition](#safe-assignment-in-condition).
   ````
 
 * <a name="no-extend-struct-new"></a>
-  Don't extend a `Struct.new` - it already is a new class. Extending it
-  introduces a superfluous class level and may also introduce weird errors if
-  the file is required multiple times.
+  Don't extend instance initialized by `Struct.new`. Extending it introduces
```
bbatsov: an instance

README.md
```diff
@@ -2301,11 +2301,21 @@ condition](#safe-assignment-in-condition).
...
 
+  ```Ruby
+  # bad
+  class Person < Struct.new(:first_name, :last_name)
+  end
+
+  # good
+  Person = Struct.new(:first_name, :last_name) do
+  end
+  ````
```
bbatsov: Those examples are slightly different. You should either add `do/end` to the first `Struct.new` or remove it from the second.

README.md
mmozuras: @bbatsov :+1:

README.md
mmozuras: @bbatsov removed `do/end`.

PR #408, opened 2015-01-29
Keep method arguments uniform

humzashah:
Discourage programmers to mix regular method arguments with the new keyword method arguments, especially where default method argument values are involved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3437,6 +3437,26 @@ condition](#safe-assignment-in-condition).
   initializers are exceptions for this rule).
 <sup>[[link](#no-optional-hash-params)]</sup>
 
+* <a name="uniform-keyword-arguments"></a>
+  Don't write a method that accepts regular arguments as well as keyword arguments. In addition to it being confusing, this can create a mess when default values are involved.
+
+    ```ruby
+    # bad
+    def your_method(a=nil, b: nil, c: nil)
+      [a, b, c]
+    end
+
+    # while passing a value only for the 'b' and 'c' keywords, passing value for 'a' becomes necessary:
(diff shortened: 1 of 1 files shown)


philoserf:
I like the intent. I'd prefer the commentary to emphasize consistent construction and to be stated in a position form.

humzashah:
@philoserf. if you were referring to the multi-line sing-sentence comment: it has now been moved to a single-line.

abotalov:
I disagree. Non-keyword arguments make sense if they are required. For example:
```ruby
find('#id')
find('#id', visible: false)
find('#id', disabled: false)
```

The first argument is the required `locator` and the other arguments are non-required keyword arguments.

humzashah:
@abotalov, the example I've included refers to both types of arguments having default values. As for the example you've provided, it can easily be implemented by a `opts={}`, which would still keep the nature of arguments uniform. The only use-case of mixing regular arguments type with keyword ones are for when you want to raise errors for certain keys being passed. I think this usage is rare. But even then, you could make all your arguments keywords to keep things uniform.

jdickey:
:+1:

strand:
:+1:

I think this PR could be improved by phrasing the rule as "Methods should receive standard arguments or keyword arguments, but not both. Use an option hash if you want a method with standard arguments to receive key-value pairs as an argument."

humzashah:
@strand, I like your suggestion. Unfortunately, I've deleted the forked repo as part of my Git-cleanup and from what I understand, I can no longer modify this pull-request.

strand:
Can we close this PR and continue the discussion on #435?

PR #411, opened 2015-02-03
Use to_h instead of Hash[]

kleaver:
Proposal for using `some_array.to_h` syntax instead of `Hash[some_array]`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2980,7 +2980,6 @@ condition](#safe-assignment-in-condition).
   hash = [[:foo, 'bar'], [:biz, 'baz']].to_h
   ```
 
-
 * <a name="ordered-hashes"></a>
   Rely on the fact that as of Ruby 1.9 hashes are ordered.
 <sup>[[link](#ordered-hashes)]</sup>
--- a/README.md
+++ b/README.md
@@ -2967,6 +2967,20 @@ condition](#safe-assignment-in-condition).
   email, username = data.values_at('email', 'nickname')
   ```
 
+* <a name="to_h"></a>
+  When using Ruby >= 2.0 prefer the `Array#to_h` method for converting an Array to a Hash
+  instead of `Hash.[]` syntax
+<sup>[[link](#to_h)]</sup>
+
+  ```Ruby
+  # bad
+  hash = Hash[[[:foo, 'bar'], [:biz, 'baz']]]
+
+  # good
+  hash = [[:foo, 'bar'], [:biz, 'baz']].to_h
+  ```
+
+
 * <a name="ordered-hashes"></a>
   Rely on the fact that as of Ruby 1.9 hashes are ordered.
 <sup>[[link](#ordered-hashes)]</sup>


bbatsov:
Any rationale?

onebree:
I think the rationale would be to follow `to_s`, `to_i`, and `to_a`.

bbatsov:
That's not the right way to think about this - e.g. `to_i` and `Integer` have different uses.

vedknom:
For literals, `to_h` is slower/harder to read than `Hash[]`

kleaver:
The reason I opened this PR was because of this question I asked on SO: https://stackoverflow.com/questions/28302482/hash-vs-to-h-are-they-the-same-and-if-so-which-is-better

Based on the answer given there, to me, `.to_h` just seems more predictable.  I don't know how it is slower/harder to read, I think that's just personal preference.

onebree:
From this comment alone, I do not think it should be in the style guide. Ruby 1.9.3 is still widely used, and does not include (to my knowledge) `.to_h`

https://stackoverflow.com/questions/28302482/hash-vs-to-h-are-they-the-same-and-if-so-which-is-better#comment44956310_28302482

manoevis:
But there are many with Ruby >= 2 also. So this pr sounds legit.

onebree:
Then, I think that styles included after 1.9.3 should be labeled as such. Same with styles deprecated after 1.9.3

besen:
array.to_h and Hash[] are not equivalent in all cases, I don't think this should be on the guide.

```
main » [:a, 1, :b, 2].to_h
TypeError: wrong element type Symbol at 0 (expected array)

main » Hash[:a, 1, :b, 2]
=> {:a=>1, :b=>2}
```

kleaver:
That's kind of my point is that Hash[] doesn't behave, to me, as predictably.  `to_h` forces you to pass in an array of arrays where nested arrays have only 2 elements.
```
2.2.1 :001 > Hash[[:foo,:bar], [:hello, :world]]
 => {[:foo, :bar]=>[:hello, :world]} 
2.2.1 :002 > [[:foo,:bar], [:hello, :world]].to_h
 => {:foo=>:bar, :hello=>:world}
```
And
```
2.2.1 :004 > Hash[[:foo, :bar]]
(irb):4: warning: wrong element type Symbol at 0 (expected array)
(irb):4: warning: ignoring wrong elements is deprecated, remove them explicitly
(irb):4: warning: this causes ArgumentError in the next release
(irb):4: warning: wrong element type Symbol at 1 (expected array)
(irb):4: warning: ignoring wrong elements is deprecated, remove them explicitly
(irb):4: warning: this causes ArgumentError in the next release
 => {}
2.2.1 :005 > [:foo, :bar].to_h
TypeError: wrong element type Symbol at 0 (expected array)
	from (irb):5:in `to_h'
```

onebree:
In general, I see a style guide as something that provides one syntax method over another because of popularity or such. I am probably wording this wrong, but as an example, for a standard string, both single and double quotes work the same. Why is it in the guide? It is evenly split between "doubles only for interpolation" and "doubles always". Same with padding before/after braces -- it comes down to readability/popularity, not that one way is "better" than the other.

Note: I understand there are minor benchmark differences between double and single quotes, but for the purpose of a _style_ guide, that is irrelevant.

This hash business should not be in the guide, because OP is suggesting (at initial glance) that `.to_h` and `Hash[]` are the same. They operate similarly, but as we can see in this conversation, they are fundamentally NOT the same... While `"#{ user.name }"`` is fundamentally the same as `"#{user.name}"`, and using double/single quotes (for basic strings).

kleaver:
@onebree I can see what you're saying. I guess a counter argument would be comparing it to `or` vs `||`.  They are intended to accomplish the same goal, but we don't use `or` because it doesn't behave how you would expect if you didn't know about the lower precedence of it.  I would argue that `Hash[]` doesn't behave as one would expect unless you know about it's quirks.

But, since they are fundamentally different, I think this might be one of those cases where the programmer just needs to know the difference and choose appropriately.  In addition, there may be some use cases for `Hash[]` out there that using `to_h` won't work for that we haven't discussed here.

onebree:
@kleaver :+1: I think there should be a poll, or executive decision on this, as you and I can circle around this discussion all day.

NewAlexandria:
After reading [the related discussion](http://www.benjaminoakes.com/2013/03/08/all-about-to_h-in-ruby-2/) (linked to the PR), I want to chip-in the opinion that I think the `.to_h` functionality also has inconsistent enough implementation (across the ruby versions) that it's not trivial enough for this style guide.

Even this example [from the mentioned SO Q&A](https://stackoverflow.com/questions/28302482/hash-vs-to-h-are-they-the-same-and-if-so-which-is-better) does not work in `Ruby 2.0.0-p481`
```
[[:a, 'foo'], [:b, 'bar']].to_h
#=> {:a=>"foo", :b=>"bar"}
```
instead it fails with
```
NoMethodError: undefined method `to_h' for [[:a, "foo"], [:b, "bar"]]:Array
	from (irb):115
	from /Users/newalexandria/.rbenv/versions/2.0.0-p481/bin/irb:12:in `<main>'
```

strand:
Per @onebree's suggestion, I have created a poll for this issue:

<script src="https://assets-polarb-com.a.ssl.fastly.net/assets/polar-embedded.js" async="true" data-publisher="Strand" data-poll-id="220694"></script>

strand:
Per @onebree's suggestion, I have created a poll.

<iframe src="http://strawpoll.me/embed_1/4272219" style="width: 600px; height: 390px; border: 0;">Loading poll...</iframe>

http://strawpoll.me/4272219

katgironpe:
@besen agree that this shouldn't be on the guide. Misleads people to think there is one absolutely correct way to convert an array to hash.

onebree:
@katgironpe / @besen I agree. Rather than retyping, here is my reason [from this comment](https://github.com/bbatsov/ruby-style-guide/pull/411#issuecomment-95659307)

@strand Thank you for the poll. I think you should add the option of the rule simply not being in the guide, since that is the stance I took a couple weeks ago.

katgironpe:
@onebree @strand exactly why I didn't vote.

onebree:
GOOGLE FORM: http://goo.gl/forms/O9e6kmgrUV

strand:
Drat! @onebree I'm sorry I missed that option. If you voted on the poll I created please recast your vote on the new poll.

onebree:
@strand please remove your poll link, and link to my comment with the new one, instead.

onebree:
FWIW, here are the results to my poll... The first definitely works, IDK about the second,
- [spreadsheet](https://docs.google.com/spreadsheets/d/1_ARFMT4jhCE75ZU7PjFc3TZsJauxwItjIOd_AdjJQKw/edit?usp=sharing)
- [graph](https://docs.google.com/forms/d/1qv_DObTZjf3wNoEEvBTrYhPkzxPscEwszgCmL8VGsDQ/viewanalytics?usp=form_confirm)

bbatsov:
OK, closing this.

onebree:
:+1: I totally forgot about the poll, too, as I am sure everyone else did after voting.

PR #412, opened 2015-02-05
Use Regexp.last_match(n) instead of Regexp.last_match[n]

cshaffer:
This form behaves the same as the Perl-like regex match variables by gracefully handling nil matches.

See bbatsov/rubocop/pull/1634 for more context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3177,7 +3177,7 @@ condition](#safe-assignment-in-condition).
 
 * <a name="no-perl-regexp-last-matchers"></a>
   Don't use the cryptic Perl-legacy variables denoting last regexp group
-  matches (`$1`, `$2`, etc). Use `Regexp.last_match[n]` instead.
+  matches (`$1`, `$2`, etc). Use `Regexp.last_match(n)` instead.
 <sup>[[link](#no-perl-regexp-last-matchers)]</sup>
 
   ```Ruby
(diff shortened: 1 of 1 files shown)


bbatsov:
I think you should also add a rule explaining why the use of something like `Regexp.last_match[]` is a bad idea. I doubt that many people are aware of this.

cshaffer:
I'm assuming you'd like a general rule about not accessing elements of a collection via `#[]` if an alternate accessor form is available?  Not just one about `Regexp.last_match` since there's already a rule stating that use of numbered groups should be avoided.

Or do you mean a rule from the other side? When defining an accessor method that returns a collection, you should provide an alternate form to save people from doing the `nil` check themselves, like I just blogged about [here](http://chadshaffer.com/2015/02/04/referencing-elements-and-the-superfluous-nil-check.html)?

Or both?

bbatsov:
> I'm assuming you'd like a general rule about not accessing elements of a collection via #[] if an alternate accessor form is available? Not just one about Regexp.last_match since there's already a rule stating that use of numbered groups should be avoided.

Yeah, that's exactly what I meant. Of course, we can't really know how `[]` is implemented for different classes. 

> Or do you mean a rule from the other side? When defining an accessor method that returns a collection, you should provide an alternate form to save people from doing the nil check themselves, like I just blogged about here?

Haven't thought about this, but it definitely makes sense.

cshaffer:
@bbatsov let me know if you'd prefer I split these other rules in a separate PR

bbatsov:
:+1: I'm fine with having both changes in a single PR. Thanks!

PR #418, opened 2015-03-05
Remove comment annotations

croaky:
This is the only subsection of the guide,
which is perhaps a smell.

Its recommendations conflict with
earlier recommendations in the "Comments" section:

> Write self-documenting code and ignore the rest of this section.
> Avoid writing comments to explain bad code.
> Refactor the code to make it self-explanatory.
> Keep existing comments up-to-date.
> An outdated comment is worse than no comment at all.

Each of `TODO`, `FIXME`, `OPTIMIZE`, `HACK`, and `REVIEW`
promote bad habits,
which can be better addressed by the earlier "Comments" section
or by using a project management system,
benchmarking tools,
or code review tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,6 @@ Translations of the guide are available in the following languages:
 * [Syntax](#syntax)
 * [Naming](#naming)
 * [Comments](#comments)
-  * [Comment Annotations](#comment-annotations)
 * [Classes](#classes--modules)
 * [Exceptions](#exceptions)
 * [Collections](#collections)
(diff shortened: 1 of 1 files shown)


croaky:
Just an idea! Let me know what you think. Cheers.

strand:
I like this. :+1:

NewAlexandria:
My only issue with this change is that code comment *will occur*, regardless of whether guidance for them is removed.  Semantic notes are easier to collect and report on.

strand:
@NewAlexandria I hear what you are saying, and it's certainly true that clever coders may have things in a working state, but still improvable with a code comment saying `TODO` or `OPTIMIZE`.

Would it be preferable to note that there are times where semantic notes will occur, and when they do they should follow these patterns, though when coders have a bug tracking system it is better to use that _instead of_ semantic comments?

NewAlexandria:
@strand Maybe.  Maybe for small outfits.

Bug tracking system are notorious for being the only way that upper-management knows what's going on.  In many cases, excess 'low/no priority' bugs, or similar stories, create distractions that aren't needed: a good tech manager should run reports and keep their own collection in a wiki or similar that the whole team owns.  Worthwhile things maybe get raised in retrospective or planning, and thrown on the backlog at that time.

All of this, IMO, is guidance beyond this style guide.

bbatsov:
Seems few people got involved in the discussion, so I'll close this PR. I hate the comment annotations myself, but I think at least the first few rules about styling them do belong to this document.

Issue #419, opened 2015-03-24
No guidance regarding preferential use of `(&:method)` ?


NewAlexandria:
I notice that there is no mention of the use of `(&:method)`

I see this pattern regularly, and have always thought of it as a best practice.  Can we add a section for it?

``` ruby
# bad
things.map { |thing| thing.some_method }

# good
things.map(&:some_method)

```

maasha:
Rubocop says:

    test.rb:3:1: C: Pass &:some_method as an argument to map instead of a block.
    things.map { |thing| thing.some_method }
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Which makes be believe it is already in the style guide?

philoserf:
Not all in rubocop is in the style guide. Thus only guide cops option.

https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/options.rb

maasha:
And why are not all cop actions from the style guide? I would expect that perhaps cop actions was missing, but not the other way around.

bbatsov:
> And why are not all cop actions from the style guide? I would expect that perhaps cop actions was missing, but not the other way around.

People submitted pretty sensible cops and they got merged. I forgot to update the guide accordingly.

pirj:
Covered in https://github.com/rubocop-hq/ruby-style-guide#proc-application-shorthand

PR #421, opened 2015-03-24
Add style for proc calling with simple blocks

NewAlexandria:
Add pass-as-argument style for proc invocation in simple blocks.

* [x] Removed the unrelated `block` guidance
* [x] \(hopefully) clarified guidance wording
* [x] checked to be sure that the 'bad' pattern was not used where in the guide, too

[See the discussion here](https://github.com/bbatsov/ruby-style-guide/issues/419)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1106,6 +1106,18 @@ condition](#safe-assignment-in-condition).
   'test'.upcase
   ```
 
+* <a name="single-action-blocks"></a>
+  Use the proc invocation shorthand when the invoked method is the only operation of a block.
+<sup>[[link](#single-action-blocks)]</sup>
+
+  ```Ruby
+  # bad
+  names.map { |name| name.upcase }
+
+  # good
+  names.map(&:upcase)
+  ```
+
 * <a name="single-line-blocks"></a>
   Prefer `{...}` over `do...end` for single-line blocks.  Avoid using `{...}`
   for multi-line blocks (multiline chaining is always ugly). Always use
@@ -1130,7 +1142,7 @@ condition](#safe-assignment-in-condition).
   end.map { |name| name.upcase }
 
   # good
-  names.select { |name| name.start_with?('S') }.map { |name| name.upcase }
+  names.select { |name| name.start_with?('S') }.map(&:upcase)
   ```
 
   Some will argue that multiline chaining would look OK with the use of {...},


NewAlexandria:
Right, so awkwardly enough, if you're doing a puts then you do need an each:

```Ruby
names.each { |name| puts name }
```

because `puts` is not a method of any element in the `names` array.  This only works when invoking the same method (or returning the same property) of each object in the array / collection.

README.md
```diff
@@ -1124,13 +1124,25 @@ condition](#safe-assignment-in-condition).
...
+  ```
 
+* <a name="block-chaining">Chaining</a>
+
+  ```Ruby
   # good
-  names.select { |name| name.start_with?('S') }.map { |name| name.upcase }
+  names.select { |name| name.start_with?("S") }.map(&:upcase)
+
+  # bad
```
bbatsov: Bad? This unrelated change has no place in this PR.

README.md
```diff
@@ -1124,13 +1124,25 @@ condition](#safe-assignment-in-condition).
   # good
   names.each { |name| puts name }
 
+  Proc invokation shorthand when that is the only operation of a block
```
bbatsov: This is pretty unclear. I'm OK with have a rule suggesting the use of &:something when possible, but it should have crystal clear wording and a separate bullet point.

NewAlexandria:
Pardon, it looks like I was really distracted and didn't re-review before opening.

Made changes and updated the PR description.

NewAlexandria:
I simplified this

NewAlexandria:
bump

README.md
```diff
@@ -1124,13 +1124,25 @@ condition](#safe-assignment-in-condition).
...
+  Use this proc invocation shorthand when the invoked method / attr is the only operation of a block
+
+  # bad
+  names.map { |name| name.upcase }
+
+  # good
+  names.map(&:upcase)
+  ```
+
+* <a name="block-chaining">Chaining</a>
```
bbatsov: I'm still confused by the change you made here. Using the shorthand syntax should have been a separate rule.

bbatsov:
@NewAlexandria ping

NewAlexandria:
I moved the added bit into its own rule, above.  I kept the change to the block-chaining example, which still reads well, linearly, because this rule exists above.

README.md
```diff
@@ -1106,6 +1106,19 @@ condition](#safe-assignment-in-condition).
...
+  Use the proc invocation shorthand when the invoked method / attr is the only operation of a block.
+<sup>[[link](#single-action-blocks)]</sup>
+
+  ```Ruby
+  # bad
+  names.map { |name| name.upcase }
+
+  # good
+  names.map(&:upcase)
+
```
bbatsov: Remove this blank line.

README.md
```diff
@@ -1106,6 +1106,19 @@ condition](#safe-assignment-in-condition).
   'test'.upcase
   ```
 
+* <a name="single-action-blocks"></a>
+  Use the proc invocation shorthand when the invoked method / attr is the only operation of a block.
```
bbatsov: Remove `attr` - there are only methods, nothing more.

README.md
```diff
@@ -1124,13 +1137,14 @@ condition](#safe-assignment-in-condition).
   # good
   names.each { |name| puts name }
 
+
```
bbatsov: Remove this blank.

README.md
```diff
@@ -1124,13 +1137,14 @@ condition](#safe-assignment-in-condition).
...
   # bad
   names.select do |name|
     name.start_with?('S')
   end.map { |name| name.upcase }
 
   # good
-  names.select { |name| name.start_with?('S') }.map { |name| name.upcase }
+  names.select { |name| name.start_with?("S") }.map(&:upcase)
```
bbatsov: Use single quotes around `S`.

README.md
```diff
@@ -1106,6 +1106,18 @@ condition](#safe-assignment-in-condition).
   'test'.upcase
   ```
 
+* <a name="single-action-blocks"></a>
+  Use the proc invocation shorthand when the invoked method the only operation of a block.
```
bbatsov: method is the only

PR #425, opened 2015-04-06
Don't include a trailing slash in self-closing elements.

JuanitoFatas:
HTML5 spec says they're optional.

http://dev.w3.org/html5/spec-author-view/syntax.html#syntax-start-tag

:balloon:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Prelude
 
-> Role models are important. <br/>
+> Role models are important. <br>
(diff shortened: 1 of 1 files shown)


NewAlexandria:
I didn't notice this until after merged.  I agree the existing style was better changed.  The defined form of  slash in a self-closing element includes a space, as `<br />`, so it was good to change it to one correct form or the other

PR #426, opened 2015-04-10
Change the regex percent Literals to conform to style

lpil:
The "good" example explaining when to use the percent literals syntax for regex uses `()` delimiters, while a few lines later the guide states that we should use `{}` delimiters instead.

I have changed this example to conform to the style guide.

Thanks,
Louis

### DIFF
--- a/README.md
+++ b/README.md
@@ -3345,11 +3345,11 @@ resource cleanup when possible.
 
   ```Ruby
   # bad
-  %r(\s+)
+  %r{\s+}
 
   # good
-  %r(^/(.*)$)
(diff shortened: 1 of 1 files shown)


onebree:
Other than regex sometimes containing parenthesis, is there a downside to using them over braces? I feel that using parenthesis keeps it uniform, while I associate braces with blocks, interpolation, and hashes.

bbatsov:
There's no downside.

bbatsov:
:+1: Thanks!

lpil:
Thank you. :)

PR #427, opened 2015-04-10
Improve attribute macro usage in class consistency

mikekelly:
- Recognise that attribute macros may be used in private section
- Move attribute macros to the end of a section

I think the first point is relatively uncontroversial.

The second is effectively a breaking change, at least for the use of attribute macros in the public section. Attribute macros are better off a the bottom of sections for two reasons:
1. Their behaviour is trivial and uninteresting, moving them to the bottom of the section keeps them out of the way of more interesting code.
2. It avoids the potential for accidentally leaving redundant attribute macros which get over-ridden by subsequent methods defined on the class

### DIFF
--- a/README.md
+++ b/README.md
@@ -2069,10 +2069,7 @@ condition](#safe-assignment-in-condition).
 
     # constants are next
     SOME_CONSTANT = 20
-
-    # afterwards we have attribute macros
-    attr_reader :name
-
+    
     # followed by other macros (if any)
     validates :name
 
@@ -2083,6 +2080,9 @@ condition](#safe-assignment-in-condition).
     # followed by public instance methods
     def some_method
     end
+    
+    # finishing off with public attribute macros
+    attr_reader :public_attribute
 
     # protected and private methods are grouped near the end
     protected
@@ -2094,6 +2094,9 @@ condition](#safe-assignment-in-condition).
 
     def some_private_method
     end
+    
+    # finally we have private attribute macros
+    attr_reader :private_attribute
   end
   ```
 


bbatsov:
> Their behaviour is trivial and uninteresting, moving them to the bottom of the section keeps them out of the way of more interesting code.

and makes them easier to miss

> It avoids the potential for accidentally leaving redundant attribute macros which get over-ridden by subsequent methods defined on the class

Seems to me this should be a concern of lint tools.

mikekelly:
> and makes them easier to miss

Well, something has to be at the bottom of the class... my point was that it makes most sense for that to be trivial stuff like attr macros.

> Seems to me this should be a concern of lint tools.

Perhaps. Regardless, this is an objective benefit of the style.

strand:
:-1: I'm used to seeing `attr_*` before `def` defined methods.

mikekelly:
@strand me too. But the question is why? In other words; does moving them under present any problems or benefits?

strand:
Attributes are the properties of an object. Reading them first is preferable, as knowing what properties an object holds quickly give the reader a sense of the _aboutness_ of an object. Thus prefer them sooner rather than later in the class.

mikekelly:
The distinction between a "property" and any other public method is completely arbitrary. You could very well write `attr_reader :thing` as

```
def thing
  @thing
end
```

The only reason you would not do that is for the sake of brevity. Not because `#thing` is in some special category of the object's interface (public or private).

mikekelly:
In fact, that is a good way to think about this...

If you wrote all of your `attr_*` declarations by hand as normal method definitions (ie. using `def` instead), where would you put them? I would definitely stick them at the bottom of either the public or private section. You?

gustavolobo:
This makes so much sense that I don't understand why wasn't it merged already #goodpractices

strand:
If I were writing `attr`s by hand I would put them in alphabetical order with other methods, unless there were a more compelling code organization strategy being employed specified in our style guide.

`attr` at the top is very common in most codebases. This PR should not be merged, as it does not reflect current community practices or practices which we have gathered a consensus that the style provides any advantage.

@mikekelly @bbatsov I recommend you close this PR.

gustavolobo:
@strand I thought this repo was also a good way of teaching the community about some good practices. Don't think that "code in this way just because the community does so" is a good argument for closing this PR.

`attr`s are not the properties of an object as we have instance variables for that purpose. `attr`s are just a nicer way to write getters and setters.

In fact, the current state of this repo is misguiding people to unnecessarily expose public methods that should be private, and that's maybe why you see so many codebases doing it wrong.

bbatsov:
> @strand I thought this repo was also a good way of teaching the community about some good practices. Don't think that "code in this way just because the community does so" is a good argument for closing this PR.

We also have to respect the strongly established conventions in the community. I understand the points you're trying to make, but it's clear that not everyone's excited about them.

gustavolobo:
I always come back to this thread and laugh a bit more about the last comment.

> Let's establish the wrong example, because the community is not excited about the good one.

😂

mikekelly:
Ha, I completely forgot about this. Good times.

I still think closing this issue is a mistake. The guide should nudge people towards objectively better practices. There's nothing in this thread that refutes the two points in the OP.

bbatsov:
Well, just think about how many people would have started complaining if that was merged and RuboCop started enforcing it by default. ;-)

gustavolobo:
And here we have another awesome reason to not implement a good standard.

mikekelly:
@bbatsov the severity could be warning, it can be turned off via simple config change - that is not a good argument at all. Any chance you could address the benefits in the OP?

bbatsov:
There's also the philosophical question of whether something name "attribute", which is generally a public property of an object, should be private in the first place. 

I'm obviously all for marking private parts of an API as private, but I'm also quite bother that you can have a private attribute. Generally I'd either roll out a private reader method direction or I won't even bother with it and just access the underlying instance var directly (as obviously a generated reader won't be doing anything special with it). 

I think the real solution here would be a better explanation of the concept of an attribute. WDYT?

mikekelly:
That's not really a philosophical question, you're just asking for "attribute" to mean more than it does. If it was meant to mean what you suggest, the language would prevent them from existing within protected or private blocks.

Writing it out by hand when you can use a language construct to avoid doing it doesn't make sense unless we're living by this made up rule that "attributes" are always public (which they aren't, as above).

Accessing an instance variable directly is sub-optimal because it makes the code more difficult to change in the future ie. with attributes if you decide you want to refactor from a simple `attr_reader` to method that returns a computed value, none of the calling code would need to change. It also means you have less redundant `@`s everywhere in your code, so you're saving wasted chars and improving readability.

To be honest, I get the impression that your primary reason for resisting this is not down to reason but because of your personal ties to the-way-you-do-things. I've clearly explained why this is a good idea in multiple reasoned ways, and you haven't directly addressed any of them.

I hope you're able to do the ruby world a favour and incorporate this change. It's a style that helped me in my Ruby career... and based on feedback, other people who've work with my code too.

bbatsov:
> To be honest, I get the impression that your primary reason for resisting this is not down to reason but because of your personal ties to the-way-you-do-things. I've clearly explained why this is a good idea in multiple reasoned ways, and you haven't directly addressed any of them.

I really don't understand why you try to make everything personal here. Absolutely everything you said is open to debate and interpretation just like half of programming. If everything was so clear cut we wouldn't be having this conversation to begin with. In this particular conversation I see two people who feel the suggestion is a good idea and two people who are not convinced.

My main concern are not any personal preferences of mine, but the fact that I've almost never seen the suggested style out in the wild and this can certainly the handled differently as well. 

> Accessing an instance variable directly is sub-optimal because it makes the code more difficult to change in the future ie. with attributes if you decide you want to refactor from a simple attr_reader to method that returns a computed value, none of the calling code would need to change. It also means you have less redundant @s everywhere in your code, so you're saving wasted chars and improving readability.

Case and point here - if you use `attr_reader` (for instance) then you'll have to replace this usage with a custom one anyways, if you end up in this situation. 😉Anyways, I'm not saying "no", that's a bad idea, I'm just saying that I'd like to see more opinions and stronger arguments.

mikekelly:
> I really don't understand why you try to make everything personal here. Absolutely everything you said is open to debate and interpretation just like half of programming. If everything was so clear cut we wouldn't be having this conversation to begin with. In this particular conversation I see two people who feel the suggestion is a good idea and two people who are not convinced

I think I've laid out some fairly objective benefits, and tried to avoid just making the case that it's just what I prefer. I still can't see anywhere you've addressed any of these points or raised concrete objections apart from a vague reference to "philosophy" about what an "attribute" is which doesn't correspond with how `attr_reader` actually works in practice.

> Case and point here - if you use attr_reader (for instance) then you'll have to replace this usage with a custom one anyways, if you end up in this situation.

I think you missed the point - I was addressing the _client_ code calling the `attr_reader`, ie. if you code is littered with references to `@thing` rather than the method `thing`, then all of these need to change if/when you decide to make the `thing` method more complex than just returning the `@thing`.

We also haven't addressed the point that putting `attr_reader` at the top means that if you did over-ride it with a method definition further down, you could easily leave it redundantly without realising it - which make reading the top of the code where you see the now-redundant `attr_reader` completely misleading.

A lot of your objection seems to be based on an unnecessary distinction between `attr_reader` and a method. An `attr_reader` is a method definition written in shorthand - they're functionally exactly the same thing.

Issue #370, opened 2014-09-24
Slightly contradicting information on embedded expressions in strings


glacials:
In the [spaces-operators](https://github.com/bbatsov/ruby-style-guide#spaces-operators) section, the two options for embedded expressions in strings are clearly presented:

> As far as embedded expressions go, there are also two acceptable options:
> ```ruby
> # good - no spaces
> "string#{expr}"
> 
> # ok - arguably more readable
> "string#{ expr }"
> ```
> The first style is extremely more popular and you're generally advised to stick with it. The second, on the other hand, is (arguably) a bit more readable. As with hashes - pick one style and apply it consistently.

But in the [string-interpolation](https://github.com/bbatsov/ruby-style-guide#pad-string-interpolation) section, a line reads:

> Consider padding string interpolation code with space. It more clearly sets the code apart from the string.
> ```ruby
> "#{ user.last_name }, #{ user.first_name }"
> ```

without presenting the alternative. This second section seems more one-sided and doesn't line up with the "choose your own" recommendation from the spaces-operators section. It also seems a bit floppy for a style guide -- "consider" isn't a word that conveys any strong convention.

In addition to these two points, it's just odd that this subject is mentioned in two different places of the guide, neither with a reference to the other.

onebree:
@glacials If you have not done so already, this weekend I will open up a pull request with an edit to the style guide.

glacials:
I have not, go for it.

dideler:
Can this be closed now?

glacials:
Yes, seems fixed!

PR #429, opened 2015-04-16
include 2 methods of spacing string interpolation

onebree:
As per issue #370 I am including 2 padding methods for sting interpolation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,8 +227,8 @@ Translations of the guide are available in the following languages:
   ```
 
   `{` and `}` deserve a bit of clarification, since they are used
-  for block and hash literals, as well as embedded expressions in
-  strings. For hash literals two styles are considered acceptable.
+  for block and hash literals, as well as string interpolation.
+  For hash literals two styles are considered acceptable.
 
   ```Ruby
   # good - space after { and before }
@@ -242,22 +242,6 @@ Translations of the guide are available in the following languages:
   popular in the Ruby community in general). The second variant has
   the advantage of adding visual difference between block and hash
   literals. Whichever one you pick - apply it consistently.
-
-  As far as embedded expressions go, there are also two acceptable
-  options:
-
-  ```Ruby
-  # good - no spaces
-  "string#{expr}"
-
-  # ok - arguably more readable
-  "string#{ expr }"
-  ```
-
-  The first style is extremely more popular and you're generally
-  advised to stick with it. The second, on the other hand, is
-  (arguably) a bit more readable. As with hashes - pick one style
-  and apply it consistently.
 
 * <a name="no-spaces-braces"></a>
   No spaces after `(`, `[` or before `]`, `)`.
@@ -3046,13 +3030,16 @@ resource cleanup when possible.
   email_with_name = format('%s <%s>', user.name, user.email)
   ```
 
-* <a name="pad-string-interpolation"></a>
-  Consider padding string interpolation code with space. It more clearly sets
-  the code apart from the string.
+* <a name="string-interpolation"></a>
+  With interpolated expressions, there should be no padded-spacing inside the braces.
 <sup>[[link](#pad-string-interpolation)]</sup>
 
   ```Ruby
-  "#{ user.last_name }, #{ user.first_name }"
+  # bad
+  "From: #{ user.first_name }, #{ user.last_name }"
+  
+  # good
+  "From: #{user.first_name}, #{user.last_name}"
   ```
 
 * <a name="consistent-string-literals"></a>


bbatsov:
The second variant is used pretty rarely in practice, which is the reason it never got included in the guide.

onebree:
If you mean the second as `"#{ user.name }`", THAT was the part in the guide. I normally see the variant I added (without space-padding)

philoserf:
:-1:

bbatsov:
> If you mean the second as "#{ user.name }", THAT was the part in the guide. I normally see the variant I added (without space-padding)

Ah, I totally missed this. I don't even remember when this rule got in the guide.

onebree:
If the rule with added-padding is not used much by the community, then it should be removed from both sections of the README.

I can modify my PR to what you prefer.

bbatsov:
Guess that'd be best, unless there are strong objections for this.

onebree:
What if I change the wording as `good` and `okay - not preferable`? Especially since there was a downvote earlier today.

bbatsov:
You can replace the first two sentences with something saying that interpolated expressions should have no spaces around them and add a note after the examples saying that adding a single space is an option, albeit not a popular one.

onebree:
I just modified the strong interpolation section. Should I also edit embedded expressions where there is padded-spacing?

If there is one, I do not see a difference between string interpolation and embedded expressions.

bbatsov:
Oh, yeah. There's none - just remove this section about embedded expressions. Guess it was placed there to be close to the rules about `{}`.

README.md
```diff
@@ -244,7 +244,7 @@ Translations of the guide are available in the following languages:
   literals. Whichever one you pick - apply it consistently.
 
   As far as embedded expressions go, there are also two acceptable
-  options:
+  options
```
bbatsov: Why did you remove this?

onebree: I did not notice that. Must have hit backspace while reading that section. As you said in https://github.com/bbatsov/ruby-style-guide/pull/429#issuecomment-95178829 I will remove the section in general.

bbatsov: OK

README.md
```diff
@@ -3047,13 +3031,18 @@ resource cleanup when possible.
   ```
 
 * <a name="pad-string-interpolation"></a>
```
bbatsov: This should be named simply `string-interpolation`.

README.md
```diff
@@ -227,8 +227,8 @@ Translations of the guide are available in the following languages:
   ```
 
   `{` and `}` deserve a bit of clarification, since they are used
-  for block and hash literals, as well as embedded expressions in
-  strings. For hash literals two styles are considered acceptable.
+  for block and hash literals, as well as string interpolation.
```
bbatsov: Remove the trailing space here.

README.md
```diff
@@ -3047,13 +3031,18 @@ resource cleanup when possible.
...
-  "#{ user.last_name }, #{ user.first_name }"
+  # good - no spaces
+  "From: #{user.first_name}, #{user.last_name}"
+ 
+  # okay - not recommended, yet arguably easier to read
+  "From: #{ user.first_name }, #{ user.last_name }"
   ```
+  
+  As in the second example, adding a _single_ space after `{` and before `}` is an option, albeit not a popular one. Like always, stick to one style throughout your code.
+<sup>[[link](#pad-string-interpolation)]</sup>
```
bbatsov: The link should be in the beginning of the rule.

philoserf:
I am not in favor of this "Okay" version. If folks want to deviate from the community norms then go for it. The norm clearly is no spaces.

Speaking in the same language as found in the okay of the most recent read me edit, "Like, just do want you want if you aren't following the standard but don't polute the standard with uncommon cases."

I remain :-1:

philoserf:
To clarify a bit more.  I think an "Okay" section is warranted when the community has demonstrated two competing styles regularly. If, on the other hand, the community has consistently used a style an "Okay" should not be included. Do we really want to promote a minority opinion? Such is counter, in my view, to the meaning of _standard_.

onebree:
@philoserf Thank you for the input. I agree. If supported by @bbatsov I will remove the "okay" option. Would I then need the example at all? (Since it would be self-explanatory, to me at least).

I kept the "okay" example because of the original purpose for the PR -- there were two styles in the guide, and I wanted to address that. But I agree that pretty much everyone omits the added space padding.

Again, let me reference issue #370, which shows this problem in the guide.

philoserf:
@onebree Thank you for the reference. I did need to re-read it. I believe the space/padding language should removed from the guide. It is both rare and only seems to serve to muddy the matter. 

What say yee, followers of issue #370 and this PR?

bbatsov:
I'm fine with just labelling the padded version as bad and be done with it.

README.md
```diff
@@ -3046,14 +3030,19 @@ resource cleanup when possible.
...
   ```Ruby
-  "#{ user.last_name }, #{ user.first_name }"
+  # bad - with spaces as padding
+  "From: #{ user.first_name }, #{ user.last_name }"
+  
+  # good
+  "From: #{user.first_name}, #{user.last_name}"
   ```
+  
+  As in the second example, adding a _single_ space after `{` and before `}` is an option, albeit not a popular one. Like always, stick to one style throughout your code.
```
bbatsov: This sentence has to be removed.

bbatsov:
@onebree Remove the sentence after examples and squash the commits into one. Let's wrap this finally. Please, capitalise the final commit message.

onebree:
Ok, will do. I will get this done by the weekend -- I need to learn rebasing. Why the strict guidelines for commits/messages, though? Just to keep the history uncluttered?

bbatsov:
> Just to keep the history uncluttered?

Yep. 5 commits for a small change doesn't make sense to me. Most commits are WIP and have no place in the commit history.

onebree:
Hi there. The date for the rebase was set as today. Hope that is fine.. Everything should be good now!

bbatsov:
:+1: Thanks!

BM5k:
:-1: for not including the WILDLY MORE READABLE `” { 1 + 1 }”` as an option. Yet another cop that will just always be disabled. :rage:

bbatsov:
@BM5k The styles supported by cops are not bound by the guide. I'd gladly accept a PR implementing the alternative style. Whether it's WILDLY MORE READLE is pretty subjective. You seem to think so, some people will definitely disagree.

onebree:
@BM5k I think you mean: `"#{ 1 + 1 }"`?

I would like to say that my original commits (before squashing) did include a method for padding around curly braces. After the discussion in this PR, it was decided to just include what is in the guide right now.

I think the guide is meant for the common/better practices in Ruby, where RuboCop is more flexible in letting you express your own style. As this discussion shows, using spaces is rare, meaning the guide would show the community-accepted one. But since it is still acceptable in terms of syntax, it makes sense for the RuboCop alternative.

PR #431, opened 2015-04-27
Fix inconsistency for an array of strings

d-unseductable:
Fix an example of single-line blocks rule.

Replace an array of strings with literal syntax `%w()` to match [this rule](https://github.com/bbatsov/ruby-style-guide#percent-w)


bbatsov:
:+1:

PR #434, opened 2015-05-03
Add initialize to consistent class structure.

strand:
In #272 @josephjaber noted the consistent class structure lacks specification where the `initialize` definition should be placed. @nilbus found that it is commonly the first instance method definition.

This change addends the consistent class structure so `initialize` is defined before other instance method definitions.


bbatsov:
:+1:  No objections here. Thanks!

PR #435, opened 2015-05-03
Use either keyword or standard arguments in method signatures.

strand:
In #408 @humzashah proposed specifying that keyword and standard arguments not be mixed, but they are no longer maintaining that PR branch.

This change would specify to use either keyword or standard arguments, but not both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3500,6 +3500,37 @@ resource cleanup when possible.
   initializers are exceptions for this rule).
 <sup>[[link](#no-optional-hash-params)]</sup>
 
+* <a name="uniform-keyword-arguments"></a>
+  Methods should receive standard arguments or keyword arguments, but not both. Use an option hash if you want a method with standard arguments to receive key-value pairs as an argument.
+
+    ```ruby
+    # bad
+    def your_method(a=nil, b: nil, c: nil)
+      [a, b, c]
+    end
+
+    # while passing a value only for the 'b' and 'c' keywords,
+    # passing value for 'a' becomes necessary:
+    your_method(nil, c: 1)
+
+    # good–using keyword arguments
+    def your_method(a: nil, b: nil, c: nil)
+      [a, b, c]
+    end
+    your_method(c: 1)
+
+    # good–using standard arguments, commonly used for initialization
+    def your_method(opts={})
+      a = opts[:a] || nil
+      b = opts[:b] || nil
+      c = opts[:c] || nil
+
+      [a, b, c]
+    end
+    your_method(c: 1)
+  ```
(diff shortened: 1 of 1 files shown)


abotalov:
IMO it's fine if a method receives standard arguments without default values and keyword arguments. It's just not fine to use standard arguments with default values.

humzashah:
@abotalov, to clarify you mean the standard+keywords arguments mixture is
fine as long as the standard arguments aren't using default values? I agree
with that. And I suspect so does @strand. Which is why the PR gives an
example of the "bad" case.

But which one of the following would you present to a new Rubyist:

1. Use either standard-args or keyword-args - don't mix them.
2. Use any args-style you like. But don't mix standard args with keyword
args when you are setting default values for some/all standard args being
used.

For me, it would be the first one - simple, easy to remember, and easy to
follow.

Regards,
Syed Humza Shah

abotalov:
I believe 2 is better than 1.

Mixing standard args with default values and keyword args is bad as it would lead to the necessity to always pass value for a standard arg with "default" value if you also want to pass keyword arg.

I don't see drawbacks in mixing standard args without default values and keyword args. I even believe it's a better code style than mixing standard args and options hash.

Keyword args have some advantages over options hash:
* no need to validate hash keys inside the method (API user may mistype them)
* no need to write boilerplate code to extract hash options
* no need to read the body of the method to understand which arguments it takes

humzashah:
@abotalov, evidently whether to stick with one style of arguments is a debatable suggestion. what we can agree on however is to not mix the argument styles when any of the standard arguments is accepting a default value.

@strand, perhaps it would be a good idea to change the PR to that (while still referencing the original PR of course). i think it would attract more widespread acceptance than the current form.

bbatsov:
I think that instead of trying to fit all definition into one of two categories we should be thinking of cases which are bad beyond any doubt. I don't have anything against mixing positional and keyword arguments - this is a pattern common in many languages.

humzashah:
I think there will be unanimous agreement that using standards arguments with default values in conjunction with keyword arguments is bad.

bbatsov:
> I think there will be unanimous agreement that using standards arguments with default values in conjunction with keyword arguments is bad.

Yep, this seems pretty likely.

strand:
It's unclear to me what, if any, consensus there is about the acceptability of mixing standard and keyword arguments. This is not surprising as keyword arguments are a new language feature. @bbatsov, do you have any specific recommendations? If not I'll keep this PR open, until we can get a clear consensus on this.

strand:
I think that the community consensus is to not mix standard and keyword arguments. @bbatsov do you agree?

pirj:
Closing as stale.

PR #438, opened 2015-05-07
Avoid defining methods inside of methods.

### LINKED ISSUE #437: Add a note about nested function declaration
Unfortunately sometimes I see something like this
```ruby
def foo(x)
  def bar(y)
    do_work y
  end
  bar(x)
end
```
in the wild and it's author doesn't know that it isn't nested function declaration, but full-blown function declaration (and re-declaration on each foo() call).

I think that it won't hurt if there will be mention of this with suggestion to use lambda/Proc or "non-nested" function declaration.

strand:
fixes #437 

Instead use lambdas to define processes at a local scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,31 @@ Translations of the guide are available in the following languages:
   end
   ```
 
+
+* <a name="method-definition-inside-methods"></a>
+    Avoid defining methods inside of methods, as this dynamically defines the "internal" method at the same scope as the "external" method. Instead use a lambda to define a process local to the scope of the method.
+  <sup>[[link](#method-definition-inside-methods)]</sup>
+
+    ```Ruby
+    # good
+    def some_method(argument)
+      internal_process = lambda do |foobar|
+        do_work foobar
+      end
+
+      internal_process(argument)
+    end
+
+    # bad
+    def some_method(argument)
+      def another_method(foobar)
+        do_work foobar
+      end
+
+      another_method(argument)
+    end
+    ```
+
 * <a name="no-trailing-params-comma"></a>
   Avoid comma after the last parameter in a method call, especially when the
   parameters are not on separate lines.


README.md
```diff
@@ -402,6 +402,31 @@ Translations of the guide are available in the following languages:
   end
   ```
 
+
+* <a name="method-definition-inside-methods"></a>
+    Avoid defining methods inside of methods, as this dynamically defines the "internal" method at the same scope as the "external" method. Instead use a lambda to define a process local to the scope of the method.
```
bbatsov: internal -> nested
external -> outer
process?

README.md
```diff
@@ -402,6 +402,31 @@ Translations of the guide are available in the following languages:
   end
   ```
 
+
+* <a name="method-definition-inside-methods"></a>
```
bbatsov: nested-method-definition perhaps?

README.md
```diff
@@ -402,6 +402,31 @@ Translations of the guide are available in the following languages:
...
+    Avoid defining methods inside of methods, as this dynamically defines the "internal" method at the same scope as the "external" method. Instead use a lambda to define a process local to the scope of the method.
+  <sup>[[link](#method-definition-inside-methods)]</sup>
+
+    ```Ruby
+    # good
+    def some_method(argument)
+      internal_process = lambda do |foobar|
```
bbatsov: process doesn't sound right to me

bbatsov:
Using code examples within a class might make explaining this even easier - http://stackoverflow.com/a/4865161/291550

ojab:
I'm thinking about something like [this](https://github.com/ojab/ruby-style-guide/commit/018091e134ff336488e23ade0c450c7f855f71d1).
IMHO guide should show the actual scope of inner method definition.

No sure if proc should be mentioned though, since it's semantic differs from method/lambda.

bbatsov:
@ojab Overall your suggestion looks good. I wouldn't mention the proc.

ojab:
Hopefully fixed. Should I make PR or my changes will be incorporated into this one? I'm ok with both options.

strand:
Hey @ojab, Just saw all the feedback. I'll integrate it and amend the commit. I'll avoid the word process, rename the anchor, use a class, and use nested and outer instead of internal and external.

strand:
Oh, @ojab I just saw that you've written a commit on this already. Submit a PR and I will close mine.

ojab:
ok, made PR https://github.com/bbatsov/ruby-style-guide/pull/439

PR #440, opened 2015-05-12
Adding a section on parallel assignment

rrosenblum:
As discussed in the RuboCop pull request for parallel assignment, https://github.com/bbatsov/rubocop/pull/1875.

I wasn't sure where exactly to put this. Let me know if it needs tweaking or moving.

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,6 +635,49 @@ Translations of the guide are available in the following languages:
    end
    ```
 
+* <a name="parallel-assignment"></a>
+    Avoid the use of parallel assignment for defining variables. Parallel
+    assignment is allowed when it is the return of a method call, used with
+    the splat operator, or when used to swap variable assignment. Parallel
+    assignment is less readable than separate assignment. It is also slightly
+    slower than separate assignment.
+<sup>[[link](#parallel-assignment)]</sup>
+
+  ```Ruby
+  # bad
+  a, b, c, d = 'foo', 'bar', 'baz', 'foobar'
+
+  # good
+  a = 'foo'
+  b = 'bar'
+  c = 'baz'
+  d = 'foobar'
+
+  # good - swapping variable assignment
+  # Swapping variable assignment is a special case because it will allow you to
+  # swap the values that are assigned to each variable.
+  a = 'foo'
+  b = 'bar'
+
+  a, b = b, a
+  puts a # => 'bar'
+  puts b # => 'foo'
+
+  # good - method return
+  def multi_return
+    [1, 2]
+  end
+
+  first, second = multi_return
+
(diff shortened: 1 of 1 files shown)


bbatsov:
A small remark about the commit message - Adding -> Add.

README.md
```diff
@@ -635,6 +635,43 @@ Translations of the guide are available in the following languages:
    end
    ```
 
+* <a name="parallel-assignment"></a>
+    Avoid the use of parallel assignment for defining variables, except for
```
bbatsov: Maybe you can break down this long sentence into two sentences?

README.md
```diff
@@ -635,6 +635,43 @@ Translations of the guide are available in the following languages:
...
+    the return of method calls, use with the splat operator, or variable
+    reassignment. Parallel assignment is less readable and slower.
```
bbatsov: `and a bit slower` - let's not put heavy emphasis on this

README.md
```diff
@@ -635,6 +635,43 @@ Translations of the guide are available in the following languages:
...
+  # bad
+  a, b, c, d = 'foo', 'bar', 'baz', 'foobar'
+
+  # good
+  a = 'foo'
+  b = 'bar'
+  c = 'baz'
+  d = 'foobar'
+
+  # good - variable reassignment
```
bbatsov: Maybe we should go into a bit more details why this particular reassignment is good (as there are other cases which won't be - e.g. `a, b = 1, 2; a, b = 3, 4`.

rrosenblum: I agree, having some extra information about variable reassignment would be good. I added a sub section on variable reassignment. I think that it may need to be rewritten though.

README.md
```diff
@@ -635,6 +635,47 @@ Translations of the guide are available in the following languages:
...
    ```
 
+* <a name="parallel-assignment"></a>
+    Avoid the use of parallel assignment for defining variables. Parallel
+    assignment is allowed when it is the return of a method call, used with
+    the splat operator, or when used for variable reassignment. Parallel
+    assignment is less readable than separate assignment. It is also slightly
+    slower than separate assignment.
+<sup>[[link](#parallel-assignment)]</sup>
+ *  Variable reasignment is acceptable because it allows you to swap values
```
bbatsov: I'd just move this into the code examples as a longer comment.

rrosenblum:
I have changed the naming of variable reassignment to variable swapping. I think that this better defines what is being done (I plan on making the same naming change inside of RuboCop). I have also moved the explanation to the comments section. Let me know what you think of the changes.

bbatsov:
:+1: Looks good. Thanks!

PR #443, opened 2015-05-12
Corrected broken links.

arbox:


bbatsov:
:+1: 10x

onebree:
Thanks for that, @arbox -- I used `git rebase` for the first time and missed that change.

arbox:
You are welcome, @onebree

0x01f7:
As far as I know, this PR did not fix the problem. You should use **pad-string-interpolation** rather than **string-interpolation**, since the href name **string-interpolation** had been already token by the 1st rule in section **Strings**.

```Markdown
## Strings

* <a name="string-interpolation"></a>
...
<sup>[[link](#string-interpolation)]</sup>

...

* <a name="pad-string-interpolation"></a>
...
<sup>[[link](#pad-string-interpolation)]</sup>
```

PR #445, opened 2015-05-14
Add options for separating method definitions

kbparagua:
Generally, we are using _only_ one empty line to separate paragraphs inside methods/blocks. I believe that code readability will increase if we will not use the same separator for methods (_group of paragraphs_), maybe we can use 2 empty lines to make a much obvious separation for methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,10 +367,9 @@ Translations of the guide are available in the following languages:
     end
   ```
 
-* <a name="empty-lines-between-methods"></a>
-  Use empty lines between method definitions and also to break up a method
-  into logical paragraphs internally.
-<sup>[[link](#empty-lines-between-methods)]</sup>
+* <a name="empty-lines-between-method-paragraphs"></a>
+  Use empty lines to break up a method into logical paragraphs internally.
+<sup>[[link](#empty-lines-between-method-paragraphs)]</sup>
 
   ```Ruby
   def some_method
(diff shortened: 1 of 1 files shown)


bbatsov:
As one class should ideally do just one thing I can't imagine how its methods won't be closely related. I'm against this suggestion, but I'll leave this PR open for a while to see if others want to discuss it.

kbparagua:
@bbatsov I agree with you that methods inside a class should be closely related. I think the 2-empty-lines option is to just make a clearer visual cue that tells the reader that the method already ends. I think the idea is almost similar to indentation, which gives the reader a visual cue that he is entering/leaving a scope.

onebree:
I understand how this makes code more readable, but I agree -- methods defined in the same class are related. Double new-line not only add excess LOC, but separates the code too much. I think double spacing would be good for tests, like contexts maybe.

For me, a single line between methods are enough, because vim highlights `def`/`end` nicely. Indentation is more crucial as it actually helps tracking excess `end` and figuring out the flow of the method (`if`, blocks, etc). Whereas this PR would just provide sugar to a program.

A way to kind of double-space between methods is to write a comment above a definition line, but that is superfluous -- method names should, in most cases, explain themselves.

bbatsov:
Seems there's no support for the proposed changes.

PR #446, opened 2015-05-23
singleton -> class change in code example

maicher:
Fix the lack of consistency between rule and code example.


bbatsov:
:+1: Thanks for spotting this!

PR #451, opened 2015-06-12
Prefer `Time` over `DateTime`

andyw8:


mentero:
Someone was reading: https://gist.github.com/pixeltrix/e2298822dd89d854444b ;)

bbatsov:
Generally it'd be better if you inlined the examples. Referring to external sources is a bad idea as they tend to change/disappear (and you force the reader to change context).

bbatsov:
@andyw8 ping

andyw8:
I'll come back to this when I have time.

PR #454, opened 2015-06-21
Fix Spelling mistake and change CamelCase into PascalCase

JamesWatling:
Hi, I have fixed a small spelling mistake, and I believe that we should use the term PascalCase instead of camelCase for Classes/Modules.

This is because camelCase (in most programming languages) generally starts with a lowercase letter. The examples show that we obviously want to start with a capital, which is the exact usage of PascalCase.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,7 +392,7 @@ Translations of the guide are available in the following languages:
 <sup>[[link](#no-trailing-params-comma)]</sup>
 
   ```Ruby
-  # bad - easier to move/add/remove parameters, but still not preferred
+  # bad - easier to move/add/remove parameters, but stil not preferred
   some_method(
                size,
                count,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -392,7 +392,7 @@ Translations of the guide are available in the following languages:
 <sup>[[link](#no-trailing-params-comma)]</sup>
 
   ```Ruby
-  # bad - easier to move/add/remove parameters, but still not preferred
+  # bad - easier to move/add/remove parameters, but stil not preferred
```
bbatsov: You're change is actually incorrect. :-)

README.md
```diff
@@ -1895,7 +1895,7 @@ no parameters.
   ```
 
 * <a name="camelcase-classes"></a>
-  Use `CamelCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
+  Use `PascalCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
```
bbatsov: I'm aware of the term `PascalCase`, but I'm pretty sure most people are not - therefore the decision to use `CamelCase` (although it's actually `camelCase`).

onebree:
When I see rules on class and module naming for Ruby, it's usually referenced as "upper camel case," letting someone know that the first letter is capitalized as well.

I've never heard of Pascal case myself. I think it's important for the guide to use terms already familiar to everyone, since the focus is best practices in Ruby.

Issue #455, opened 2015-06-23
Use RDoc and its conventions for API documentation.


maxcal:
I don't really think this guide needs to be opinionated on RDoc vs YARD vs Whatever. The main point is that your should document your API's and while doing so follow the conventions of the documentation tool.

philoserf:
:+1: @maxcal

jdickey:
:+1: to that as well. RDoc was adequate in the days when JavaDoc was new and shiny; the world *has* moved on.

bbatsov:
Yeah, I'm fine with dropping this and replacing it with some generic guidelines about API documentation.

PR #456, opened 2015-06-30
Reword the bit about documentation to not sound as RDoc biased. Fixes…

maxcal:
… #455

### DIFF
--- a/README.md
+++ b/README.md
@@ -554,9 +554,8 @@ Translations of the guide are available in the following languages:
   num = 1_000_000
   ```
 
-* <a name="rdoc-conventions"></a>
-    Use RDoc and its conventions for API documentation.  Don't put an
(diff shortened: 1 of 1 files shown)


bbatsov:
Or we can just kill this item altogether. Doesn't make much sense in it's current form. Perhaps just suggesting to document APIs would be enough.

maxcal:
Ok :)

lorentrogers:
Is there no standard-practice around documenting each function? What should the documentation include? Should all methods be documented? Should private methods be documented? 

These were all questions I had while reading through the Rubocop source: https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/file_name.rb

I had assumed that each method should have at least *some* comment on it's general purpose and the arguments it accepts? Is this not the case?

PR #465, opened 2015-08-07
correct code example of optional arguments

kangkyu:
This way it keeps the code give out the result in example, I guess

### DIFF
--- a/README.md
+++ b/README.md
@@ -657,7 +657,7 @@ Translations of the guide are available in the following languages:
   some_method('w', 'x', 'y', 'z') # => 'w, x, y, z'
 
   # good
-  def some_method(c, d, a = 1, b = 2)
+  def some_method(a, b, c = 1, d = 2)
     puts "#{a}, #{b}, #{c}, #{d}"
   end
 


mentero:
:+1: nice catch

arbox:
I don't think this correction should be accepted since it breaks the compatibility with the previous bad example.

mentero:
But the current one doesn't work. Output from the examples is just wrong.

arbox:
I think we'd rather correct the output, but up to the editor :)

onebree:
I can go either way with this PR. I see why have `c` and `d` the predefined arguments, but at the same time, I think it depends on how the variables are called. I like having the arguments in order as to how I call them, so if `a` is called first in the method AND is predefined, I will place it last `bcda` in the argument list.

kangkyu:
Anyways it should be corrected. By the way how about options hash instead.

bbatsov:
Let's correct the output.

kangkyu:
Output has been corrected https://github.com/bbatsov/ruby-style-guide/commit/31b229bacb9e1bc510412ca7628e0b79eba2449e this pull request can be closed 😄

PR #467, opened 2015-08-12
Don't rescue Exception in exception-ordering rule

etdev:
The ``#good`` code example for the [exception-ordering](https://github.com/bbatsov/ruby-style-guide#exception-ordering) rule rescues from ``Exception``, which we're specifically told to avoid in the previous rule, [no-blind-rescues](https://github.com/bbatsov/ruby-style-guide#no-blind-rescues).  

I propose switching to ``IOError`` and ``StandardError`` in the [exception-ordering](https://github.com/bbatsov/ruby-style-guide#exception-ordering) rule.

Also removes unnecessary newline from [no-blind-rescues](https://github.com/bbatsov/ruby-style-guide#no-blind-rescues).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2857,7 +2857,6 @@ no parameters.
   # also good
   begin
     # an exception occurs here
-
   rescue StandardError => e
     # exception handling
   end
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: Fine by me.

PR #468, opened 2015-08-12
Add correct usage ex. for percent-q rule

etdev:
Add an example of when to use the ``%q()`` operator to [percent-q](https://github.com/bbatsov/ruby-style-guide/blob/master/README.md#percent-q) rule.


bbatsov:
:+1:

PR #469, opened 2015-08-12
Add example code for gsub-blocks

etdev:
Adds example usages of `sub`/`gsub` with block and hash arguments to the [gsub-blocks](https://github.com/bbatsov/ruby-style-guide/blob/master/README.md#gsub-blocks) rule

Also fixes grammar slightly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3439,8 +3439,14 @@ resource cleanup when possible.
   ```
 
 * <a name="gsub-blocks"></a>
-  For complex replacements `sub`/`gsub` can be used with block or hash.
+  For complex replacements `sub`/`gsub` can be used with a block or a hash.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3439,9 +3439,15 @@ resource cleanup when possible.
   ```
 
 * <a name="gsub-blocks"></a>
-  For complex replacements `sub`/`gsub` can be used with block or hash.
+  For complex replacements `sub`/`gsub` can be used with a block or hash.
```
bbatsov: or a hash

etdev: You're right, thanks!  I've updated the file to make it ``a block or a hash``.

onebree:
I think this is very helpful. :+1:

PR #470, opened 2015-08-12
Re-word non-capturing-regexp rule

etdev:
Fixes slightly awkward wording for [non-capturing-regexp](https://github.com/bbatsov/ruby-style-guide/blob/master/README.md#non-capturing-regexp) rule

Also moves ``# bad`` and ``# good` comments to the line above their corresponding code examples, in line with the other code examples.


bbatsov:
:+1: Slightly awkward? You're too kind. :-)

PR #471, opened 2015-08-12
Add #bad example for string-concat

etdev:
Adds `#bad` example for [concat-strings](https://github.com/bbatsov/ruby-style-guide/blob/master/README.md#concat-strings) rule.


bbatsov:
:+1:

PR #474, opened 2015-08-13
Escape characters for proper syntax highlighting

yous:
This makes Vim happy. :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,11 +103,11 @@ Translations of the guide are available in the following languages:
 
 * <a name="utf-8"></a>
   Use `UTF-8` as the source file encoding.
-<sup>[[link](#utf-8)]</sup>
+<sup>\[[link](#utf-8)\]</sup>
 
 * <a name="spaces-indentation"></a>
   Use two **spaces** per indentation level (aka soft tabs). No hard tabs.
-<sup>[[link](#spaces-indentation)]</sup>
+<sup>\[[link](#spaces-indentation)\]</sup>
 
   ```Ruby
   # bad - four spaces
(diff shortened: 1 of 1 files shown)


bbatsov:
This escaping is definitely not necessary, so I'd rather us not add it as typing those links is already enough of a chore.

yous:
Then how about replacing `[`, `]` with `(`, `)`? Parentheses don't need escape.

bbatsov:
It's all the same to me, although square brackets obviously don't require escaping, otherwise this wouldn't be working. And this is something I use in other projects as well, if it's changed here I'll have to change it everywhere, which is not something I'm looking to deal with.

yous:
Sorry, but what do you mean "otherwise this wouldn't be working"? Actually `[` and `]` around `[link](...)` need escaping just for editors because editors try to parse a link, but `(` and `)` around `[link](...)` don't need escaping for editors.

bbatsov:
> editors try to parse a link

Editors or vim? :-) If this was a problem in Emacs I would have noticed it by now.

> "otherwise this wouldn't be working"

If escaping was really needed, the markdown processors wouldn't be able to handle this, which is clearly not the case.

yous:
Alright. I understood your case. Maybe I should find some better markdown plugin. Closing.

onebree:
@yous FWIW, I have vanilla vim on my Koding.com VM. It definitely treats `md` files differently than `markdown`. I suggest finding a plugin or editing your `.vimrc` to treat `md` just as `markdown`. If you try changing the extension on your branch, you may see it "properly" highlighted.

Also, note that the only highlighting this guide is concerned with is such for the ruby code blocks. As long as the markdown works, anything else is eye candy

yous:
Hi @onebree, actually I'm using `autocmd BufNewFile,BufRead *.md setfiletype markdown` script to make `.md` files be recognized as markdown. Additionally I'm using [tpope/vim-markdown](https://github.com/tpope/vim-markdown), but it seems to have some issues with syntax highlighting with GitHub flavored markdown.

PR #475, opened 2015-08-21
Use the strictest type conversions possible

nathanl:


onebree:
My only suggestion is to reword it so it is easier to understand for those newer to Ruby. I am familiar with these methods, but I did not understand the explanation until the example... That, or I am not paying enough attention.

bbatsov:
@nathanl I see your point and I tend to agree, although I've seen much bigger usage of `.to_i` & friends in my experience. I also feel that from time to time it's reasonable to use `.to_i` and friends to dispose of some bad data, but in general I'd prefer your approach.

I wonder what others would say on the subject.

nathanl:
@bbatsov Yeah, there are definitely times when looser conversion is OK, but eg, when reading from a spreadsheet and adding up numbers, if you accidentally read from the product description column instead of the product price column, it's much better to get `ArgumentError: invalid value for Integer(): "pickle juicer"` than "everything's cool, looks like your total price is 0!"

I guess the guideline is really like "be as clear as possible about what you expect". It's the same reason I like `hash.fetch` - it either makes it clear that I *really* need this value to be there, or forces me to clarify what to do if it isn't by giving a default. Either way, someone reading the code understands my intention, and someone getting an error knows why.

@onebree Thanks for the feedback. I'll take another stab at making it understandable if you all agree that it's a good guideline.

jdickey:
@nathani, @bbatsov I've recently had to work on some code where the convention was "use strict conversions, and conditionally recover with a looser conversion if configuration allows." Simulated example:

```ruby
    # assume some_input is 'foo'
    begin
      some_existing_var = Integer(some_input, 10)
    rescue ArgumentError => e
      raise unless some_config_variable
      logger.info "caught error #{e.message}; coercing some_existing_var to 0 per config"
      some_existing_var = 0
    end
```

> It seemed like a good idea at the time&hellip;

PR #481, opened 2015-09-09
Remove `Foo = Class.new(...)`

LavirtheWhiolet:
Let we have `Foo = Class.new(Bar)`.
1. Imagine that we need to add a method into Foo.
2. It is not uniform (with other class declarations, `class Foo < Bar`).
3. It is cryptic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,19 +155,21 @@ Translations of the guide are available in the following languages:
   ```
 
 * <a name="single-line-classes"></a>
-  Prefer a single-line format for class definitions with no body.
+  Do not add empty lines in class definitions with no body.
 <sup>[[link](#single-line-classes)]</sup>
 
   ```Ruby
   # bad
   class FooError < StandardError
+  
   end
-
-  # okish
-  class FooError < StandardError; end
-
+  
   # good
-  FooError = Class.new(StandardError)
+  class FooError < StandardError
+  end
+  
+  # good (but not preferable)
+  class FooError < StandardError; end
   ```
 
 * <a name="no-single-line-methods"></a>


LavirtheWhiolet:
Maybe one should add an advice to use metaprogramming for oneline method-less classes.

bbatsov:
Let's hear what others have to say about this. I see your points, although the general idea is to use the shorthand only when it's really unlikely to add anything to the class going forward.

lsegal:
+1. I will add that documentation tools will not be able to reliably pick-up the `Foo = Class.new` form. From that perspective, I agree that this is cryptic. I wouldn't recommend this form to devs unless they had a good reason to break from the standard practice of using the `class` keyword. There are a few reasons where this is needed, but none of them should fit into a style guide (rather, their usages don't violate any other styles).

A single lined `class X; end` is easier to parse both from the eyes of a human reader and an automated tool.

bbatsov:
Well, I guess we can move the `Class.new` variant to bad and add some rationale against its usage. In general it'd would be nice if we added a rule about when it makes sense to use `Class.new` over a regular class definition.

kluzny:
Also, when I go to expand the class and want to actually add a method, I end up reverting this format back to the traditional class definition. This also breaks the searchability of class definitions `ack 'class.*Error'`

bbatsov:
Any updates here?

bbatsov:
I'm closing this PR due to no recent activity. Feel free to re-open it if you ever come back to it.

marcandre:
Sorry I'm late to the party, but I wholeheartedly agree with the PR 👍

I personally can not see what is wrong with this form:

```ruby
class Foo
end
```

It can simply not be clearer. 100% of human readers will parse this instantly.
It's also the minimum keystrokes to type.
It can be refactored the easiest.

It's not possible to miss the `; end` after a long declaration:
```ruby
class SomethingOrRather < ::RuboCop::Cop::SomethingElse::MyBaseClass; end

module Foo # is this correctly indented? Oh, yes it is, I missed the "; end"
# ...
```

Let's not forget, we write code for it [to be easy to read](https://www.goodreads.com/quotes/835238-indeed-the-ratio-of-time-spent-reading-versus-writing-is)

bbatsov:
@marcandre Feel free to wrap it up. I closed it only due to the lack of activity.

PR #483, opened 2015-09-22
Add a section about trailing underscore variables

rrosenblum:


aredondo:
Why prefixing `foo` with the splat operator? If `foo` is an array, don't the following two assignments work equally:

    *a, b, _ = *foo
    *a, b, _ = foo

bbatsov:
> If foo is an array, don't the following two assignments work equally:

There are other splattable objects, for this won't be the case.

README.md
```diff
@@ -708,6 +708,28 @@ Translations of the guide are available in the following languages:
   a = *(1..3)
   ```
 
+* <a name="trailing-underscore-variables"></a>
+  Avoid the use of unnecessary trailing underscore variables during
+  parallel assignment. Trailing underscore variables are necessary
+  when there is a splat variable defined in the argument chain, and
```
bbatsov: in the argument chain -> on the left side of the assignment (sounds better to me)

rrosenblum:
Updated, let me know if this sounds better.

0x01f7:
It's really hard to understand, since we know nothing about **foo**, so I open an [issue](https://github.com/bbatsov/ruby-style-guide/issues/504), colud have a look?

The section is about **unnecessary trailing underscore variables**, why not give a meaning about **foo**? And I still cannot agree the meaning about **unnecessary trailing underscore variable** as I asked in Q4.

Sorry for my poor English.

rrosenblum:
You make an excellent point. Take a look at #506. I have attempted to make the variables more meaningful and provide a bit of reasoning.

PR #484, opened 2015-09-24
Indent and align long hashes

gnilrets:


bbatsov:
I'm not sure if we need a rule for this - for me it's apparent that's the way to go.

gnilrets:
@bbatsov, I don't think it's apparent to everyone.  I've definitely seen long, complex hashes written out on a single line.

onebree:
https://github.com/bbatsov/ruby-style-guide#no-double-indent

If you ask me, key/value pairs in a hash can be considered _parameters_ of the hash object. So this PR would be redundant.

pirj:
This can improve readability in some cases. However, we allow method calls with keyword arguments to be all on the same line:
```ruby
Taxonomy.detect(name: 'Certainly Citrus', abv: 7.3, hops: { flavoring: 'Citra', bittering: 'Cascade'})
```

and it strongly depends on a number of factors (line length limit, nesting) on when it reaches the point when it's hard to read on a single line.

The same, I believe, applies to hash literals. This makes it hard to come up with a rule that is specific to hash literals.

To me, the current https://github.com/rubocop-hq/ruby-style-guide#no-double-indent does not specify if it is possible to put several semantically grouped arguments on one line:
```ruby
# good or bad?
User.create(
  first_name: 'Bob', last_name: 'Doe',
  zip: '78 ux 32', city: 'New Dum',
  interests: [:poker, :aikido, :meditation])
```

Closing, as there is no clear way to define this guideline.

PR #485, opened 2015-09-29
salary in Russian is "зарплата"

solutus:


solutus:
sorry, my mistake. Bulgarian and Russian are close, but not similar =)

PR #486, opened 2015-10-06
Add a section about nested modifiers

lumeet:


README.md
```diff
@@ -968,6 +968,19 @@ Translations of the guide are available in the following languages:
   end
   ```
 
+* <a name="no-nested-modifiers"></a>
+  Avoid nested modifier `if/unless/while/until` usage. Favor `&&/||` if it is
```
onebree: _if appropriate_, I see what you did there.

Issue #487, opened 2015-10-09
inline private and protected


itsNikolay:
What about inline `private` and `protected` method declaration?
For example:
```ruby
def perform
  do_it
  do_then
end

private def do_it
  # action
end

private def do_then
  # action
end
```
Does it increase readability?

onebree:
Duplicate of #449

bbatsov:
The community doesn't have a unified stance on this yet.

Issue #449, opened 2015-06-04
Recommendation for inline private, protected method


timcolonel:
I was wondering if there was a recommendation on the way of defining private/protected methods since ruby 2.1

```ruby
# Method 1
private

def foo

end

# Method 2
private def foo

end
```

onebree:
How has it changed since Ruby 2.1? I am on Ruby 1.9.3 still for work. We define private methods in classes by listing `private` on its own line, then it's followed by the method definitions.

timcolonel:
Well i am pretty sure ruby 2.1 made method definition return a symbol which can be passed to private/protected and it was not working before 
* https://bugs.ruby-lang.org/issues/3753: Feature ticket
* http://www.sitepoint.com/look-ruby-2-1/: Section 2

onebree:
Does saying `private` only work for the immediate method definition, or for all following ones?

timcolonel:
If you put `private` alone then all the following methods will be private but if you but the definition on the same line `private def ...` then only this method will be private.

```ruby
class Test
  def foo
    'foo'
  end

  private def bar
    'bar'
  end

  def pub
    'pub'
  end
end


t = Test.new
puts t.foo
puts t.pub
puts t.bar
```
Will print
```
foo
pub
Uncaught exception: private method `bar' called for #<Test:0x30aea38>
```

onebree:
Okay that makes sense. Thank you very much.

equivalent:
in contrast how about "indented" private/ protected https://github.com/bbatsov/ruby-style-guide/pull/488 ?

bbatsov:
The community doesn't seem to have a strong position regarding this, so I'd rather not add anything to the guide at this time. Maybe a few years down the line.

dmnd:
It's now been 18 months. Is the community ready for being able to tell when a method is private without scrolling up?

khiav223577:
How about
```rb
# Method 3
private \
def private_method
  ...
end
```

Enether:
@dmnd Raises a valid point. Can we not enforce our own strong position?

eliav-lavi:
@dmnd, I feel that the inline `private` declaration encourages longer class, where the Ruby mindset would tend to prefer shorter classes where scrolling is not an issue (or is a minor issue).

maria-lau:
@bbatsov now that it's been a few years, can we revisit this topic?

bbatsov:
[...]

PR #488, opened 2015-10-09
Private and Protected methods should have extra level of indentation

equivalent:
This will solve the issue of having too many private methods and loosing
visual track of which methods are public and which private, therefore
doing inline `private` 
- https://github.com/bbatsov/ruby-style-guide/issues/449
- https://github.com/bbatsov/ruby-style-guide/issues/487

This also comply to scissors rule http://www.eq8.eu/blogs/16-scissors-rule-in-coding

### DIFF
--- a/README.md
+++ b/README.md
@@ -2262,16 +2262,14 @@ no parameters.
     def some_method
     end
 
-    # protected and private methods are grouped near the end
+    # protected and private methods are grouped near the end indented with extra level
     protected
-
-    def some_protected_method
-    end
+      def some_protected_method
+      end
 
     private
-
-    def some_private_method
-    end
+      def some_private_method
+      end
   end
   ```
 
@@ -2593,14 +2591,13 @@ no parameters.
     end
 
     private
+      def private_method
+        # ...
+      end
 
-    def private_method
-      # ...
-    end
-
-    def another_private_method
-      # ...
-    end
+      def another_private_method
+        # ...
+      end
   end
   ```
 


equivalent:
the "file changed" tab comparison is not really showing my point; it should look like this:

```ruby
class Foo
 
  def blabla
  end

  protected
    def fobar
    end

  private
    def woohoo
    end
```

onebree:
I would like to note, FWIW, throughout the Rails Guides, they follow the proposed convention.

equivalent:
@onebree yep, yep `:)`   and in my experience is really more pragmatic concept, when working with larger projects (Applications, large lib instead of smaller code base lib) 

So far only disagreement  I came across to not to use indented private was:

**A:** it's confusing when you have multiple classes in one file to fallow which level belongs to which class

```ruby
module Foo
   module Bar
      module Car
        class Tom
          class Jhon        
            def jhon1
            end
             
            private
               def pjhon1
               end

               def pjhon2
               end
        end

        def tom
        end

        def tom2
        end

        private
           def ptom3
           end

           def ptom4
           end
    end
  end
end
```

... but if developer end's up with this, he is already doing it wrong (should split to own classes) :smile: 

**B:** Ruby Style Guide is showing this in same indent level :)

jdickey:
All I can say is that this *better* default to disabled, because it's going to break probably the majority of code in actual production with *absolutely zero* real definable benefit that I can see.

* #449 is a classic "how does this work" question that exposes two *different* and reasonably complementary uses of the `private` keyword. Since there are *technical, semantic* meanings between the two styles quoted in the initial comment for that issue, a style gude/enforcer should have *no opinion* on its use unless we develop technology to read the mind of the developer as to which he meant &mdash; technology that I have yet to see a working PR for, on this or any other project;
* #487 is a simple restatement of a single case impliled by #449;

The "scissors rule" blog post is one guy's opinion that most effectively demonstrates how critical being able to communicate effectively is to software development, again completely independent of any technology used.

Let's kill this thing (or at least default it to `disabled`) before it encourages a mass emigration away from the tool&hellip;which I see as entirely plausible given similar experiences in multiple other languages. Let's focus on effectively communicating a shared understanding of the code &mdash; for which a style guide and tool *can be* helpful up to but *not exceeding* the point where they jump any and all available sharks. This, in my view, is just such an ichthyological acrobatic feat.

etdev:
I don't really like this because
* If you have so many private methods you can't keep track of where they start/end, you should probably refactor the class anyway.  Recommending this method implies that it's okay to get yourself into this situation.
* It wastes two spaces unnecessarily, decreasing the amount of horizontal space you have for actual code.
* It can get confusing when you have inner classes etc.

equivalent:
hi @etdev, with all respect but:

> it can get confusing when you have inner classes

and 

> If you have so many private methods you can't keep track of where they start/end, you should probably refactor the class anyway

are conflicting each other. 

Basically you are saying if your file has too much stuff refactor it to another file, so why would you need to have several classes in one file ? (that is already covered by this guide " Don't nest multi line classes within classes")

in tearms of:

> It wastes two spaces unnecessarily, decreasing the amount of horizontal space you have for actual code.

yes that is true. But if we take that as valid argument not to implement it, then  the same argument can be used on  "indent nested class inside module". So therefore this is not valid:

```ruby
module Foo
  module Bar
    class Car
    end
  end
end
```

...and this is :

```ruby
module Foo
module Bar
class Car
end
end
end
```

double edge sword :smiley: 

But I'm not trying to be ignorant here. I definitely see that in case of private/protected we are nesting below keyword, where in module/class we are nesting inside kind of block (class-end, module-end) scope. I'm just saying that this point is not good enough argument for myself, but maybe that's just me

jdickey:
We have several situations where we use nested classes, but each class has its implementation and tests in files separate from those in the namespaces "above" or "below" it (or other closely-related classes at the same level and therefore in the same namespace).

An example from my current [research project](https://github.com/jdickey/new_fortstrap/); we have a class, [`Views::Welcome::Index::PostSummary::Body`](https://github.com/jdickey/new_fortstrap/blob/59b7279/app/views/welcome/index/post_summary/body.rb), with a class definition that starts out

```ruby

module Views
  module Welcome
    class Index < Views::Base
      # Widget representing a summary view of a Post instance.
      class PostSummary < Views::Base
        # Widget encapsulating representation of post-summary body content.
        class Body < Views::Base
        # ... Magic Happens Here ...
        end
      end
    end
  end
end
```

* `Views` and `Views::Welcome` are modules used (mostly) as namespaces (the outer is required for the [view-template replacement Gem](https://github.com/ageweke/fortitude)), so nothing special there;
* `Views::Welcome::Index` is a class in its own right, with its own [code](https://github.com/jdickey/new_fortstrap/blob/59b7279/app/views/welcome/index.rb) and [test](https://github.com/jdickey/new_fortstrap/blob/59b7279/test/views/welcome/index_test.rb) files;
* `Views::Welcome::Index::PostSummary` is a class that, although used exclusively by `Welcome::Index`, has its own separate [code](https://github.com/jdickey/new_fortstrap/blob/59b7279/app/views/welcome/index/post_summary.rb) and [test](https://github.com/jdickey/new_fortstrap/blob/59b7279/test/views/welcome/index/post_summary_test.rb) files;
* The `Views::Welcome::Index::PostSummary::Body` class, excerpted above, likewise is used exclusively by `Views::Welcome::Index::PostSummary`, yet has its own code and test files like any other class in the app.

> But aren't you introducing coupling with all that nesting?

Yes, but only to the specific degree and in the specific manner supported by our conventions. If a class is used *exclusively* to support another class, then its naming/namespacing should reflect that as an aid to rapid perusal and navigation. If a class is more general-purpose, it will be moved to a different namespace, and the nesting redefined accordingly. Ruby doesn't give us the tools (AFAIK) to make that easier but, in most cases, that's a feature, not a bug.

> What about SOLID design? Doesn't your hierarchy defeat that? What about DRY?

Not really. Our tools, and the conventions we've built or bought into around those tools, provide strong encouragement for the single-responsibility principle, *both at the method and class levels.* Each class should have only one reason to change *within the context of its implementation logic.* Therefore, our view is that having long, deeply-coupled methods within a class is almost always an SRP violation, but simply renaming (and re-positioning) a class in the overall hierarchy is a *packaging issue*, and thus *does not* violate SRP.

The open-closed principle can be addressed by having small, single-purpose classes that inject anything likely to change, such as configuration or styling settings. The use of blocks in the form of injected procs is our currently-favoured mechanism for explicitly opening seams at points where code, not just data, may need to change.

The LSP, ISP, and DRY are taken into account by defining relatively generic classes for concepts that might have several very tightly-related (and highly similar) variations. We define such general-purpose classes using dependency injection to address the configuration data and "seam joints"  where changes are likely, and then one or more layers of aggregation around those generic classes that themselves expose increasingly more specific, simpler interfaces.

Finally (for SOLID), dependency inversion is a classic use for Ruby blocks of various species.

Anything I've missed?

> When you believe you're wielding a sword with only two edges, either it is a thing of overwhelmingly elegant, sophisticated simplicity or, more likely, you're soon to discover new and completely unpredicted ways to slice yourself to ribbons.

equivalent:
I'm sorry @jdickey I have problem fallowing your thoughts here (and I mean that for both comments) It's not that I don't understand what are you trying to say it's just it appears to you are talking about different topic.

Sorry that I won't be able to reply to any of your comments.

Can you please summarize your points against this PR ? (where every point is less than 10 words, so I understand your clear intention)


As for nesting class inside class. I've never said anything against it, it' just that code should be in own file. So `View::Welcome::Index::PostSummary::Body`

```ruby
module Views
  module Welcome
    class Index < Views::Base
      class PostSummary < Views::Base
        class Body < Views::Base
           # code 
        end
      end
    end
  end
end
```

....is perfectly valid, as long as: `View::Welcome::Index` and `View::Welcome::Index::PostSummary` have their own files where the code is implemented.

cainlevy:
I find that indenting for protected and private methods does not improve my ability to tell whether any single method is non-public. It's ambiguous with the indentation from nesting inside a Ruby module. So there is no increase in readability compared to not indenting. In both cases, the clues necessary to tell the difference are somewhere else.

fuadsaud:
Isn't order good enough to structure visibility? https://github.com/bbatsov/ruby-style-guide#consistent-classes

bbatsov:
Closing due to lack of enough support.

Issue #415, opened 2015-02-11
`and` / `or` keywords with `raise` / `fail`


smanolloff:
The idea of banning the `and` & `or` keywords is really cool.

However, there is one specific case in which they may be preferred:

``` ruby
authenticate_user! || fail AuthenticationFailure, 'Invalid credentials'
# SyntaxError: unexpected '\n', expecting :: or '[' or '.'

authenticate_user! or fail AuthenticationFailure, 'Invalid credentials'
# (this works ok)
```

Of course, we we could just use

``` ruby
fail AuthenticationFailure, 'Invalid credentials' unless authenticate_user!
```

... but doesn't it degrade readability?

Is the usage of `and` or `or`, immediately followed by a `raise` or `fail`, justified?
I would be happy if you could share your thoughts!

philoserf:
I find the `object.test or fail 'message'` construction is very common in the wild.

bbatsov:
> I find the object.test or fail 'message' construction is very common in the wild.

In Perl. :-) Seriously, though - yeah, it's popular to some extent but way more people use modifier `if/unless` for such code.

smanolloff:
Sure, i guess that is fine.
But we should not forget that the `if / unless` one-liners are not exactly the same as the boolean operators: 

```ruby
def current_user
  fail 'Authentication failed' unless @current_user ||= authenticate
  @current_user # need to return explicitly
end

def current_user
  @current_user ||= authenticate or fail 'Authentication failed'
end
```

luizdamim:
:+1: for allowing `and`/`or` followed by `raise`/`fail`. I use it quite frequently in my codebase.

Ajedi32:
I think the real crux of the issue is that `and` and `or` are perfectly fine to use for their intended purpose: as control flow operators. The problems they cause usually occur when people try to use them in place of `||` and `&&` as boolean operators.

Then again though, `if` and `unless` can usually be used in place of `or` and `and`, so idk...

bbatsov:
> I think the real crux of the issue is that and and or are perfectly fine to use for their intended purpose: as control flow operators.

Pretty sure those were never intended as flow of control operators. This is something which people came up with to justify their existence. It's pretty clear they were actually added to facilitate a style of programming with less parentheses in it, which is a good or a bad thing, depending who you ask.

cheerfulstoic:
Avdi Grimm has a really great (and free) RubyTapas episode about this:

[...]

Issue #244, opened 2013-11-23
Better bad examples for AND and OR


jkarttunen:
The justification for "The and and or keywords are banned. It's just not worth it. Always use && and || instead." is not very clear. The bad examples should illustrate why those should not be used better.

bbatsov:
I'm closing this ticket due to no recent activity. Feel free to re-open it if you ever come back to it.

PR #489, opened 2015-10-11
clarify why 'and' and 'or' are evil

davearonson:
This PR adds the rationale for the rule that `and` and `or` are banned.  (You may have thought it obvious, but I've seen enough people wondering, that I don't think it's as obvious as it need to be!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -914,12 +914,6 @@ Translations of the guide are available in the following languages:
   # boolean expression
   ok = got_needed_arguments and arguments_are_valid
 
-  # boolean expression
-  error = lack_needed_arguments or arguments_are_invalid
-
-  # control flow
-  document.validate and document.save
-
   # control flow
   document.save or fail(RuntimError, "Failed to save document!")
 
@@ -927,21 +921,12 @@ Translations of the guide are available in the following languages:
   # boolean expression
   ok = got_needed_arguments && arguments_are_valid
 
-  # boolean expression
-  error = lack_needed_arguments || arguments_are_invalid
-
-  # control flow
-  document.save if document.valid?  # note also more idiomatic name
-
   # control flow
   fail(RuntimeError, "Failed to save document!") unless document.save
 
   # ok
   # control flow
-  document.validate && document.save
-
-  # control flow
-  document.save || die("Failed to save document!")
+  document.save || fail(RuntimeError, "Failed to save document!")
   ```
 
 * <a name="no-multiline-ternary"></a>
--- a/README.md
+++ b/README.md
@@ -903,29 +903,56 @@ Translations of the guide are available in the following languages:
   ```
 
 * <a name="no-and-or-or"></a>
-  The `and` and `or` keywords are banned. It's just not worth it. Always use
-  `&&` and `||` instead.
+  The `and` and `or` keywords are banned. The minimal added readability is just
+  not worth the high probability of introducing subtle bugs. For boolean
+  expressions, always use `&&` and `||` instead. When doing flow control, use
+  `if` and `unless`; `&&` and `||` are also acceptable but less clear.
 <sup>[[link](#no-and-or-or)]</sup>
 
   ```Ruby
-  # bad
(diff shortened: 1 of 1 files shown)


onebree:
While this style guide is great, it is really nothing more than a style guide, providing the community's ideal way of writing ruby code. I do not see this as a tutorial guide, etc. I do not see the point of the examples you added, as they are more of a tutorial than exhibiting style.

I do agree that a small explanation is needed. Your explanation seems to just be a rewording until the last sentence.

IMO nix everything except for the sentence on lines 908 and 909.

equivalent:
I agree with the point maid by @onebree .

But If anyone ever creates a repo `ruby-best-practices-explained` I would definitely be +1  to put it there :smile:

README.md
```diff
@@ -903,30 +903,57 @@ Translations of the guide are available in the following languages:
...
+  `if` and `unless`; `&&` and `||` are also acceptable but less clear.
 <sup>[[link](#no-and-or-or)]</sup>
 
   ```Ruby
-  # bad
-  # boolean expression
-  if some_condition and some_other_condition
-    do_something
-  end
+  # BAD:
```
bbatsov: This differs from the style we employ in the document.

davearonson:
@onebree yes it is largely a rewording, so as to add the rationale.  The doc says "I've tried to add the rationale behind the rules (if it's omitted I've assumed it's pretty obvious)."  So I figured my adding the rationale for _why_ `and` and `or` are banned would be a Good Thing.  The bad-boolean examples are to back that up, providing examples of such subtle bugs.  Perhaps I did overdo it a bit on the number of examples; I will submit another revision momentarily, chopping out about half of them.

README.md
```diff
@@ -903,28 +903,30 @@ Translations of the guide are available in the following languages:
...
+  The `and` and `or` keywords are banned. The minimal added readability is just
+  not worth the high probability of introducing subtle bugs. For boolean
+  expressions, always use `&&` and `||` instead. For flow control, use
+  `if` and `unless`; `&&` and `||` are also acceptable but less clear.
 <sup>[[link](#no-and-or-or)]</sup>
 
   ```Ruby
   # bad
   # boolean expression
-  if some_condition and some_other_condition
```
bbatsov: Not sure why you opted to change the `if` examples. They are probably more common than using to booleans to set a third one.

davearonson: Because they demonstrate the subtle bugs that `and`/`or` can introduce.  In the bad one (and the other bad one I just chopped out), the variable is set to the opposite of the value that it would have been had the author used `&&`/`||`.

onebree:
I do not see it necessary to change the original examples provided for this rule. Why introduce the method `fail`, when the point of the rule is to describe `and`/`or` and `&&`/`||`?

As I mentioned before, this is a _style guide_, and not a teaching exercise or best practices guide.

@davearonson For anyone already familiar with Ruby, they know that `and`/`or` has lower precedence and can be an issue when combining with `&&`/`||`. Knowing this, why risk an error in the code? _Just use what you know will work._ (The latter option)

dpisarewski:
I'm tired of people wanting to ban these keywords. You just should not use `and`/`or` with non-logical operators and assignment, or mix them with `&&`/`||`. All other cases are just fine.

```
#bad
#logical expression with assignment
variable = yay or nay

#mixing and with ||
if ruby and java || scala
  do_something
end

#mixing logical and bitwise operations
flag = 1 or 0b0011 & 0b10001 

#good
#assignment with ||
variable = yay || nay

#logical expression
if ruby and (java or scala)
  do_something
end
```

dpisarewski:
Here are other discussions on the same issue:

https://github.com/bbatsov/ruby-style-guide/issues/244
https://github.com/bbatsov/ruby-style-guide/issues/415
https://github.com/bbatsov/ruby-style-guide/pull/295
https://github.com/bbatsov/ruby-style-guide/commit/5920497452c1f6f604742a735f5684e86d4c0003

PR #490, opened 2015-10-13
fix examples with optional arguments

AlexanderPavlenko:
tested in 2.2.2p95


bbatsov:
Thanks for spotting this - was fixed in another PR.

PR #491, opened 2015-10-13
Clean up wording slightly

etdev:
* Remove first-person statement about people disagreeing with `case-when` indication rule, as it's superfluous information that doesn't really add anything to the guide.
* Change `method` to `methods` to bring it in line with the plurality of the rest of the sentence


bbatsov:
Thanks!

PR #494, opened 2015-10-31
Feedback request - Proc Immediate Invocation - which style do you favor?

eliotsykes:
In researching the Ruby equivalent of IIFEs in JavaScript, there didn't seem to be any documented conventions on immediately invoking procs. 

In this PR, you'll see 3 variations for immediately invoked procs. 

Feedback welcome: https://github.com/eliotsykes/ruby-style-guide/blob/proc-immediate-invoke/README.md#proc-immediate-invoke

### DIFF
--- a/README.md
+++ b/README.md
@@ -1627,6 +1627,22 @@ no parameters.
   l.call(1)
   ```
 
+* <a name="proc-immediate-invoke"></a>
+  Prefer newline/inline immediate invocation of procs.
+<sup>[[link](#proc-immediate-invoke)]</sup>
+
+  ```Ruby
+  # newline - preferable or not?
+  l = -> { something }
+  l.call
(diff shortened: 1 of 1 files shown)


eliotsykes:
Here's usage in the wild to aid discussion:

```ruby
module JsonHelper
  def json
    (reparse_and_never_memoize_as_response_may_change = -> do
      JSON.parse(response.body, symbolize_names: true)
    end).call
  end

  def json_request_headers
    { 'Content-Type' => 'application/json', 'Accept' => 'application/json' }
  end
end

RSpec.configure do |config|
  config.include JsonHelper, type: :request
end

```

bbatsov:
I always go with option 1. It's the clearest as far as I'm concerned.

PR #498, opened 2015-11-06
Fix link to Wikipedia article on SOLID

ivan-kolmychek:
I'm sure that's because everyone knows what SOLID is and, thus, never clicked the link, but we still need to keep links correct, don't we? =)


bbatsov:
:+1:

PR #500, opened 2015-11-11
Add the namespace definition guide

maurogeorge:
Hi guys,

We have a gotcha accordingly the way we define a module[1](http://blog.elpassion.com/ruby-gotchas/).

This patch add the prefered way to declare a namespace.

This makes sense?

Thanks for the feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2700,6 +2700,29 @@ no parameters.
   end
   ```
 
+* <a name="namespace-definition"></a>
+  Use explicit nesting of modules. Definitions at a namespace are only available
+  if that namespace is defined via explicit nesting.
+<sup>[[link](#namespace-definition)]</sup>
+
+  ```Ruby
+  module API
+    URL = 'http://example.com'
+  end
+
+  # bad - does not has access to the `URL` constant
+  class API::Person
+    attr_accessor :first_name, :last_name
+  end
+
+  # good - has access to the `URL` constant
+  module API
+    class Person
+      attr_accessor :first_name, :last_name
+    end
+  end
+  ```
+
 ## Exceptions
 
 * <a name="fail-method"></a>


benissimo:
Please merge this in. I've seen too many cases of the gotcha-prone way of defining namespaces. Would love to see this preferred way become part of common practice.

Thanks to @maurogeorge for adding this!

cveneziani:
:+1:  I fully agree with both of you.

bbatsov:
The suggestion is reasonable, but I think the rationale should be extended a bit.

benissimo:
See http://stackoverflow.com/questions/15119724/ruby-lexical-scope-vs-inheritance

bbatsov:
I meant this as something that should be added to the rule. :-)

benissimo:
@bbatsov My bad, sorry I misread your earlier comment. Agree that it'd be nice to have a more explicit rationale. I've taken a stab at it in a comment on the commit. Feedback welcome!

benissimo:
p.s. one other possible rationale is that, given that there are multiple ways of specifying namespaces, it's (arguably) preferable to standardize on one way of doing so. (Pro: code is more consistent, easier to read, more familiar) Given that the explicit nesting allows for inheritance while the compressed nesting does not, it seems reasonable to standardize on the more fully featured version.

Counter-arguments:
* explicit nesting results in code that occupies more lines (less compact)
* explicit nesting results in more indentation, leaving less horizontal room per line

Those counter arguments would convince me if the two versions were functionally equivalent.
But, in my opinion, functionality trumps aesthetics. I'm not willing to renounce the possibility of inheriting constants from outer namespaces, even if it means putting up with some indentation.

maurogeorge:
Thanks for the review guys. I updated the description.

@benissimo thanks a lot for your comments, I change some things to be more concise with the guide, for example I am using code comments, like we have [here](https://github.com/bbatsov/ruby-style-guide#parallel-assignment). I prefer do not use the term constant directly on the description, because I think this behavior occurs with `module_function`, class variables etc.

bbatsov:
Closing due to lack of activity.

PR #503, opened 2015-11-18
Fix wording in #bang-not-not

etdev:
Use of the `not` keyword in this context requires parentheses `()`, not braces `{}`.


bbatsov:
:+1:

PR #506, opened 2015-11-19
Make the example for trailing underscore variables clearer

### LINKED ISSUE #504: Section `Avoid the use of unnecessary trailing underscore variables during parallel assignment` is really hard to understand.
section: [Avoid the use of unnecessary trailing underscore variables during parallel assignment](https://github.com/bbatsov/ruby-style-guide#trailing-underscore-variables)

Q1: What is the meaning about variable **foo**? method return or normal local variable?
Q2: What will we get if we call **foo**? Array, String, Range, or anything? If it's an Array, what's the size?
Q3: We will get the same result about **a**, **b** and **_**, so the splat opertor is unnecessary, right?

  ```ruby
  a, b, _ = *foo
  a, b, _ = foo
  ```

Q4: What is **unnecessary trailing underscore**?

  ```ruby
  # Starting point
  a, b, c = *foo

  # Bad? Why? I think it follows the rule: Prefix with _ unused block parameters and 
  # local variables, and much readable than `a, b, = *foo` if `foo = [1, 2, 3]`, since we
  # can immediately know the format about variable foo.
  a, b, _ = *foo 

  # Same reason here.
  a, _, _ = *foo 

  # Good? Why? You will never use _b, right? So it is unnessary, why not omit it?
  a, _b = *foo
  a, _b, = *foo
  ```

rrosenblum:
This fixes #504. 

@0x01f7 let me know what you think of the updated examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -709,24 +709,36 @@ Translations of the guide are available in the following languages:
 
 * <a name="trailing-underscore-variables"></a>
   Avoid the use of unnecessary trailing underscore variables during
-  parallel assignment. Trailing underscore variables are necessary
-  when there is a splat variable defined on the left side of the assignment,
-  and the splat variable is not an underscore.
+  parallel assignment. Named underscore variables are to be preferred over
+  underscore variables because of the context that they provide.
+  Trailing underscore variables are necessary when there is a splat variable
+  defined on the left side of the assignment, and the splat variable is
+  not an underscore.
 <sup>[[link]](#trailing-underscore-variables)</sup>
 
   ```Ruby
   # bad
-  a, b, _ = *foo
-  a, _, _ = *foo
-  a, *_ = *foo
+  foo = 'one,two,three,four,five'
+  # Unnecessary assignment that does not provide useful information
+  first, second, _ = foo.split(',')
+  first, _, _ = foo.split(',')
+  first, *_ = foo.split(',')
+
 
   # good
-  *a, _ = *foo
-  *a, b, _ = *foo
-  a, = *foo
-  a, b, = *foo
-  a, _b = *foo
-  a, _b, = *foo
+  foo = 'one,two,three,four,five'
+  # The underscores is needed to show that you want all elements
+  # except for the last number of underscore elements
+  *beginning, _ = foo.split(',')
+  *beginning, something, _ = foo.split(',')
+
+  a, = foo.split(',')
+  a, b, = foo.split(',')
+  # Unnecessary assignment to an unused variable, but the assignment
+  # provides us with useful inforation.
+  first, _second = foo.split(',')
+  first, _second, = foo.split(',')
+  first, *_ending = foo.split(',')
   ```
 
 * <a name="no-for-loops"></a>


0x01f7:
The point is the underscore variable is neither used nor providing any useful information, so omit it, right?

```ruby
foo = %w(one two three four five)

first, second, _ = foo # E1, bad
first, second, _third = foo # E2, okish, provide some useful information
first, second = foo # E3, good, omit unnessary variable
*first, second, _ = foo # E4, good, underscore is needed, first will be ['one', 'two', 'three']
```

I got it. But in my opinion, E2, E3 is still a bad style, since we could not know how many objects in **foo** unless we have read ahead or searched the API doc(if **foo** is an method return). It matters. The code below is much readable though it use unnesessary underscore variables.

```ruby
foo = %w(one two three four five)

first, second, _third, _fourth, _fifth = foo # good, no doubt
first, second, *_rest = foo # good, no doubt

fist, second, _, _, _ = foo # really bad? I think it should be good.  
first, second, *_ = foo # really bad? I think it should be good.
```

What is really bad?

```ruby
first, second, third = foo # bad, too many values to unpack(5 => 3)
first, second, third, fourth, fifth, sixth = foo # bad, not enough values to unpack(5 => 6)
```

I donnot care whether you use underscore variable or not(follow the rule [underscore-unused-vars](https://github.com/bbatsov/ruby-style-guide#underscore-unused-vars) is enough), but the count mismatch. It may a trap.

Explicit is better than implicit. Python raise a ValueError in this case. As a rubyist, I have never write code in real-world programming. What about others?

Thank you for your quick reply.

0x01f7:
In my opinion:

```ruby
foo = %w(one two three four five)

first, second, _ = foo # bad

first, second = foo # okish, extra values are ignored, 
first, second, _, _, _ = foo # good, have same count, much readable, may be?
first, second, *_ = foo # good, gather all rest element to underscore variable
```

rrosenblum:
I follow your thoughts on this. We all agree that a named underscore variable is better than an underscore variable or no variable at all. Beyond that, it partially comes down to preference. Having an underscore provides little to no context.

> But in my opinion, E2, E3 is still a bad style, since we could not know how many objects in foo unless we have read ahead or searched the API doc(if foo is an method return).

> As a rubyist, I have never write such code in real-world programming. What about others?

This is true of methods that return more than one variable, but splat throws a wrench into the mix. Lets say you have an array, but you do not know the size. `foo = [1, 2, 3, 4]` or `foo = [1, 2, 3, 4, 5]`. From this array, you want to capture the first element, the last element, and all elements that fall between. The way to write this would be `first, *middle, last = *foo`. If you didn't actually need the first or last node for some reason, an underscore could be used.

Again this comes down to preference. One of the example that you provided was `fist, second, _, _, _ = foo # really bad? I think it should be good.` The underscores provide context that this method returns 3 extra variables, but can we trust this? We don't know what these variables are either. `first, second, = foo` also tells you that this method returns extra variables. We still can't fully trust it though. Neither provides much information.

I will add a note to the style about preferring named underscore variables to named variables. 

There is a slight performance benefit to not including underscore variable or named underscore variables at all. Take a look at https://github.com/bbatsov/rubocop/pull/2246 for more info on the performance aspect.

0x01f7:
Yes. Prefer named underscore variables over underscore variables, and the underscore variables could be used unless they are necessary. Thank you add the note. I will close the issue once the PR merged. :+1:

README.md
```diff
@@ -709,24 +709,36 @@ Translations of the guide are available in the following languages:
...
+  foo = 'one,two,three,four,five'
+  # The underscores is needed to show that you want all elements
+  # except for the last number of underscore elements
+  *beginning, _ = foo.split(',')
+  *beginning, something, _ = foo.split(',')
+
+  a, = foo.split(',')
+  a, b, = foo.split(',')
+  # Unnecessary assignment to an unused variable, but the assignment
+  # provides us with useful inforation.
```
nunommc, line 738: typo in information

PR #507, opened 2015-11-23
No space around hash is not as nice

IanVaughan:


bbatsov:
Many people like it, so we won't be changing this any time soon. The community is clearly divided on this issue.

PR #513, opened 2015-12-10
Correct meaning of ||= and &&=

onebree:
It is a common misconception that `||=` means `a = a || b`. Rather, it means `a || a = b`

This change follows the official Ruby documentation's explanation of the operators.

http://ruby-doc.org/core-2.2.3/doc/syntax/assignment_rdoc.html#label-Abbreviated+Assignment

### DIFF
--- a/README.md
+++ b/README.md
@@ -1378,8 +1378,8 @@ condition](#safe-assignment-in-condition).
   x = x * y
   x = x**y
   x = x / y
-  x = x || y
-  x = x && y
+  x || x = y
+  x && x = y
 
   # good
   x += y


HorizonShadow:
It took me a second, so to explain: `a = a || b` performs an assignment operation regardless of the value of `a`. `a || a = b` only does an assignment if `a` is not truthy.

0x01f7:
As the document says, `a ||= b` behaves more like `a || a = b` than `a = a || b` **unless you have already defined `a`**. Otherwise it is not true.

onebree:
Should I just close this PR then?

Issue #512, opened 2015-12-05
Exceptions should favor `raise` over `fail`


lsegal:
TL;DR `raise` is the correct verb to communicate raising an exception in Ruby. `fail` is not a reliable method, no better semantically, and simply, should not be favored over `raise`.

First off let me say that I'm all for a great community-driven style guide, and this is a great project to run! Thank you to the people who are making this possible! That said, I believe that to have a style guide that is reinforced by the community, it should represent the community. I currently feel as though the suggested use of `fail` does not represent the Ruby community. There are stats and reasons below.

To start off, I am trying to understand where this all came from-- i.e., the rationale behind favoring `raise` over `fail` in the style guide. I've never seen much `fail` usage in my time writing Ruby. The only relevant information I can find is from a [Jim Weirich email](http://devblog.avdi.org/2014/05/21/jim-weirich-on-exceptions/) published on Avdi Grimm's blog last year. From this reading, it seems as though the rationale for `fail` was taken verbatim from Jim's suggested preference rather than any deeper technical reasoning or look at real world usage.

Although I respect Jim greatly for his many contributions to the community, I would argue strongly against trying to shoehorn this personal preference into the community as a convention. Specifically, I would argue a few major points:

1) Regardless of how you create an exception, the verbiage that we use to communicate with Ruby developers is still "*raising an Exception*", not "failing a method" or anything of the sort. By that standard, any Ruby developer that sees `raise` should clearly understand what is happening. It does not seem as clear to me that `fail` would be universally understood, especially since there *are* other semantics at play here.

2) To the "other semantics" point: `fail` will *not work* on any class inheriting BasicObject:

```ruby
class Foo < BasicObject
  def foo; fail "bye" end
end

Foo.new.foo # => NoMethodError 
```

This exemplifies a very big technical flaw in pushing for `fail`. I'm not sure why we're recommending a convention that only works in "many cases" but "not all" cases. Incidentally, this style guide *makes no mention of this pitfall*. That said, I would not want to see the solution to this problem to just add a clause about BasicObject for a few reasons: (a) a developer may not know if their class actually inherits BasicObject (consider grand-child classes), and (b) recommending a syntax that is inherently unreliable never sounds like "good style" to me. In short, I would never suggest something that "works most of the time" as a "best practice". I don't think this style guide should, either.

[...]

PR #517, opened 2015-12-17
Remove distinction between fail and raise

lee-jon:
The distinction between fail and raise isn't community adopted. Remove
the distinction and prefer raise in examples. No preference is added.

Discussion in issue:

https://github.com/bbatsov/ruby-style-guide/issues/512

### DIFF
--- a/README.md
+++ b/README.md
@@ -2714,46 +2714,44 @@ no parameters.
 
 ## Exceptions
 
-* <a name="fail-method"></a>
-  Signal exceptions using the `fail` method. Use `raise` only when catching an
-  exception and re-raising it (because here you're not failing, but explicitly
-  and purposefully raising an exception).
-<sup>[[link](#fail-method)]</sup>
+* <a name="prefer-raise"></a>
+  Prefer `raise` over `fail` for exceptions.
+  <sup>[[link](#prefer-raise-over-fail)]</sup>
 
   ```Ruby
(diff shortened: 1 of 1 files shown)


bbatsov:
Hmm, I don't think this is the way to go. The related discussion clearly suggested we discourage the use of `fail` and this is what we should do.

lee-jon:
I didn't read @lsegal comments as enforcing a preference. It as such and thought that was maybe overstepping what it was trying to achieve. Would you recommend a specific 'prefer `raise` over `fail`' rule or just remove reference to fail entirely?

bbatsov:
Prefer `raise` over `fail`.

lee-jon:
OK. Updated

README.md
```diff
@@ -2714,46 +2714,44 @@ no parameters.
 
 ## Exceptions
 
-* <a name="fail-method"></a>
-  Signal exceptions using the `fail` method. Use `raise` only when catching an
-  exception and re-raising it (because here you're not failing, but explicitly
-  and purposefully raising an exception).
-<sup>[[link](#fail-method)]</sup>
+* <a name="prefer-raise"></a>
```
0x01f7: the specified name should be `prefer-raise-over-fail` which is used later.

bbatsov: I'll fix this.

PR #519, opened 2015-12-22
changed outputs in comments of default parameter section

julioberina:


bbatsov:
Thanks for spotting it - it was fixed in another PR.

PR #524, opened 2016-01-04
Add a rule about Hash#each_key & Hash#each_value

ojab:


bbatsov:
This rule should be extended to cover cases like:

```
hash.each { |k, _v| do_something(k) }
```

ojab:
Done.

ojab:
Added case where hash is modified in block to `good` section, should it be moved to `okish` with a note that it's not a good style?

README.md
```diff
@@ -3112,6 +3112,24 @@ resource cleanup when possible.
...
+  # bad
+  hash.keys.each { |k| p k }
+  hash.values.each { |v| p v }
+  hash.each { |k, _v| p k }
+  hash.each { |_k, v| p v }
+
+  # good
+  hash.each_key { |k| p k }
+  hash.each_value { |v| p v }
+  hash.keys.each { |k| hash.delete(k) }
```
0x01f7: [Do not modify a collection while traversing it.](https://github.com/bbatsov/ruby-style-guide#no-modifying-collections)

ojab:
Whoops, reverted.

PR #527, opened 2016-01-07
Fix parameter names in good example for optional args

figureone:
Method body assumes parameters are ordered a, b, c, d, not c, d, a, b.

### DIFF
--- a/README.md
+++ b/README.md
@@ -656,7 +656,7 @@ Translations of the guide are available in the following languages:
   some_method('w', 'x', 'y', 'z') # => 'w, x, y, z'
 
   # good
-  def some_method(c, d, a = 1, b = 2)
+  def some_method(a, b, c = 1, d = 2)
     puts "#{a}, #{b}, #{c}, #{d}"
   end
 


0x01f7:
Maybe we should modify the output from the function to fix the sample, rather than the declaration?

figureone:
Not sure that would be any clearer; I would expect the output to match the order of the parameters. We could change the puts line, but I think that's just as confusing as renaming the optional parameters like in the pull request. Just my two cents!

0x01f7:
Sorry for my poor English. I mean that

```Ruby
some_method('w', 'x') # => 'w, x, 1, 2'
some_method('w', 'x', 'y') # => 'w, x, y, 2'
some_method('w', 'x', 'y', 'z') # => 'w, x, y, z'
```

should be 

```Ruby
some_method('w', 'x') # => '1, 2, w, x'
some_method('w', 'x', 'y') # => 'y, 2, w, x'
some_method('w', 'x', 'y', 'z') # => 'y, z, w, x'
```

Neither rearrange the parameter list, nor modify the method body. Because we use **a**, **b** as optinal args in **bad sample**. 

----

dup with #490 ?

figureone:
Thanks for finding the dups! In my mind it's more illustrative to keep the parameter naming and output alphabetical (a, b, c, d). This means we switch up the parameter names between good/bad examples, which is the main downside.

Both of the dups opted for leaving the un-alphabetical parameter names and fixing the example output to match. To me this confuses the point, because the output ordering doesn't match the parameter ordering.

0x01f7:
Got it. @rrosenblum, as the original author of this rule, any suggestions?

rrosenblum:
My intention was to show correlation to the previous example. Instead of declaring the optional arguments first (`a = 1` and `b = 2`), this was showing to order them last.

```ruby
def some_method(a = 1, b = 2, c, d)
  puts "#{a}, #{b}, #{c}, #{d}"
end
```
should be refactored to 
```ruby
def some_method(c, d, a = 1, b = 2)
  puts "#{a}, #{b}, #{c}, #{d}"
end
```

I can now see where this could be a bit confusing. The parameter names should be treated as important so I don't think their order should change. I would be fine changing the interior of the method from `puts "#{a}, #{b}, #{c}, #{d}"` to `puts "#{c}, #{d}, #{b}, #{a}"` and changing the outputs accordingly. Changing the interior of the method and leaving the example as is, both show that the refactor may not able to be done without consequences. In either case, it may be better to add a comment stating that there could be consequences to this change.

bbatsov:
This was finally updated.

PR #528, opened 2016-01-08
Enhanced percent literal style guide with meaningful bracket delimiters.

christianrolle:
There are 2 points I want to discuss:
1.) The style guide should point out why brackets for percent literals should be used
2.) Change the style guide in terms of using square brackets for Array literals instead of parentheses for expressiveness reasons (I think percent literal delimiters should be meaningful)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3627,21 +3627,49 @@ resource cleanup when possible.
 <sup>[[link](#percent-s)]</sup>
 
 * <a name="percent-literal-braces"></a>
-  Prefer `()` as delimiters for all `%` literals, except `%r`. Since parentheses
-  often appear inside regular expressions in many scenarios a less common
-  character like `{` might be a better choice for a delimiter, depending on the
-  regexp's content.
-<sup>[[link](#percent-literal-braces)]</sup>
+  Prefer brackets as delimiter for all `%` literals, because their usage does 
+  not need to be escaped.
 
   ```Ruby
   # bad
-  %w[one two three]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3627,22 +3627,50 @@ resource cleanup when possible.
 <sup>[[link](#percent-s)]</sup>
 
 * <a name="percent-literal-braces"></a>
-  Prefer `()` as delimiters for all `%` literals, except `%r`. Since parentheses
-  often appear inside regular expressions in many scenarios a less common
-  character like `{` might be a better choice for a delimiter, depending on the
-  regexp's content.
-<sup>[[link](#percent-literal-braces)]</sup>
+  Prefer brackets as delimiter for all `%` literals, because their usage does
```
bbatsov: Let's make the wording compatible with the previous version of the rule. E.g. you can add that `[]` delimiters are ok when you're dealing with array literals.

README.md
```diff
@@ -3627,22 +3627,50 @@ resource cleanup when possible.
...
   %w(one two three)
-  %q("Test's king!", John said.)
+  %i<one two three>
+
+  # good
+  %w[one two three]
+  %i[one two three]
   ```
 
+  and `()` for Strings and shell commands
```
bbatsov: I don't like sentences that are broken up by code examples. It makes it really hard to follow such text along.

christianrolle:
Completely reverted my pull request in favour of: #532.

PR #532, opened 2016-01-19
Enhances array literals with meaningful brackets.

christianrolle:
Change the style guide in terms of using square brackets for Array literals instead of parentheses for expressiveness reasons (I think percent literal delimiters should be meaningful). Square brackets reveal the intention, that the literal creates an Array, like:

~~~ruby
# good
%w[one two three]
%i[one two three]
~~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -3113,7 +3113,7 @@ resource cleanup when possible.
   ```
 
 * <a name="hash-each"></a>
-  Use `Hash#each_key` instead of `Hash#key.each` and `Hash#each_value`
+  Use `Hash#each_key` instead of `Hash#keys.each` and `Hash#each_value`
   instead of `Hash#values.each`.
 <sup>[[link](#hash-each)]</sup>
 
--- a/README.md
+++ b/README.md
@@ -2287,7 +2287,7 @@ no parameters.
   ```
 
 * <a name="file-classes"></a>
-  Don't nest multi line classes within classes. Try to have such nested
+  Don't nest multi-line classes within classes. Try to have such nested
   classes each in their own file in a folder named like the containing class.
 <sup>[[link](#file-classes)]</sup>
 
--- a/README.md
+++ b/README.md
@@ -2714,7 +2714,7 @@ no parameters.
 
 ## Exceptions
 
-* <a name="prefer-raise"></a>
+* <a name="prefer-raise-over-fail"></a>
   Prefer `raise` over `fail` for exceptions.
   <sup>[[link](#prefer-raise-over-fail)]</sup>
 
--- a/README.md
+++ b/README.md
@@ -3627,19 +3627,20 @@ resource cleanup when possible.
 <sup>[[link](#percent-s)]</sup>
 
 * <a name="percent-literal-braces"></a>
-  Prefer `()` as delimiters for all `%` literals, except `%r`. Since parentheses
-  often appear inside regular expressions in many scenarios a less common
-  character like `{` might be a better choice for a delimiter, depending on the
-  regexp's content.
+  Prefer `()` as delimiters for all `%` literals, except `%r` and array 
+  literals. Since parentheses often appear inside regular expressions in many 
+  scenarios a less common character like `{` might be a better choice for a 
+  delimiter, depending on the regexp's content. When creating array literals it
+  is more expressive to use `[]`.
 <sup>[[link](#percent-literal-braces)]</sup>
 
   ```Ruby
   # bad
-  %w[one two three]
+  %w<one two three>
   %q{"Test's king!", John said.}
 
   # good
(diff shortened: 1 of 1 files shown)


exAspArk:
:+1:

linchus:
:+1:

0x01f7:
I don't think it's more meaningful... :confused:

christianrolle:
What do you expect from something like "[]" to generate, compared to "()"?
In Ruby "[]" is very well know for accessing/creating a collection, don't you agree?
So why not using that practice for creating Arrays by percent literals?
I mean in Ruby we create an Array like:

~~~ruby
numbers = [1, 2, 3]
~~~

so why not leaning on it:

~~~ruby
names = %w[linchus exAspArk 0x01f7]
~~~

To me that is consistent.

README.md
```diff
@@ -3627,19 +3627,20 @@ resource cleanup when possible.
 <sup>[[link](#percent-s)]</sup>
 
 * <a name="percent-literal-braces"></a>
-  Prefer `()` as delimiters for all `%` literals, except `%r`. Since parentheses
-  often appear inside regular expressions in many scenarios a less common
-  character like `{` might be a better choice for a delimiter, depending on the
-  regexp's content.
+  Prefer `()` as delimiters for all `%` literals, except `%r` and Array
```
bbatsov: Just a tiny remark - should Array be capitalised? I guess it depends on whether we're discussing the class or arrays in general.

bbatsov:
Please change the commit message from `Enhances Array literals with meaningful brackets.` to `Enhance array literals with meaningful brackets`.

christianrolle:
good point.

christianrolle:
Done (#532).

bbatsov:
Good. Just squash those 3 commits together and kill the `.` from the title of the resulting commit to match our conventions.

0x01f7:
> What do you expect from something like "[]" to generate, compared to "()"?
> In Ruby "[]" is very well know for accessing/creating a collection, don't you agree?
> So why not using that practice for creating Arrays by percent literals?

I know it shoud be an array literal, because I see `%w`, neither `[]`, nor `()`. So both are OK for me. But use `()` as delimiters for **all `%` literals**, that is consistent to me.

Of course, In Ruby `[]` is very well know for accessing/creating a collection. But if we change the rule, what does **all `%` literals** mean?

+ %x, %s ----> avoid use
+ %r ----> use `{}`
+ %q, %Q ----> use `()`
+ %w, %W, %i, %I ----> use `[]`

I don't mind we use `()` or `[]` for array literals. But this modification make the rule unclear.

christianrolle:
Yeah. I totally agree. That's why I first wanted to remove the **all** `%` **literals** rule.
The rule already is inconsistent, since `{}` have to be preferred for regexp.

>  %r ----> use `{}`

So the Ruby style guide already does **not** suggest to use `()` for all literals.
To me a solution would be removing the recommendation to prefer `()` for all literals (since curly braces are obviously no `()`).

0x01f7:
I agree with you. The rule should be rewritten. By the way, don't forget modify the sample code in other rules([single-line-blocks](https://github.com/bbatsov/ruby-style-guide#single-line-blocks), [array-join](https://github.com/bbatsov/ruby-style-guide#array-join), [percent-w](https://github.com/bbatsov/ruby-style-guide#percent-w), etc.).

christianrolle:
Good point.
Maybe some others have some opinions about rewriting the rule as 0x01f7 and me suggest?

exAspArk:
Hi, so you are suggesting to stop using `[]` or `{}` for `%` literals? And always to use `()`?

IMHO, unifying the guide is a good idea. However, it's not so clear for me why we should prefer `()` over other characters? Is there any reasons? Why not always to use `{}` then? 
Or maybe make it not mandatory?

exAspArk:
Hi! So you are suggesting to remove the guide? :smile_cat: Or leave just a part about `{}` for `%r` literals? 

What do you think about unifying the guide? Why not to always use `{}` then?

christianrolle:
@exAspArk: I guess there is some misunderstanding. I suggest to rephrase it.
My intention was to increase expressiveness by using ´[]´ braces for array literals.

0x01f7:
> Hi! So you are suggesting to remove the guide? :smile_cat: Or leave just a part about {} for %r literals? What do you think about unifying the guide?

I suggest to
 
* Keep the original version
* Or accept this PR, but rephrase it

Both are OK for me. I am worried about the modification makes the rule too complex, and hard to understand.

> Why not to always use {} then?

Of course you can. It's all your choice.

christianrolle:
I close this pull request as not approved, since there seem to some reservations against it.

PR #539, opened 2016-02-04
Warn about scope in loop

jaredbeck:
If people blindly switch from `while` to `loop`, they may get stung by this subtle change in scope. Let's warn them.

```
while true do # exits on second iter.
  done ||= false
  break if done
  done = true
end

loop do # never exits
  done ||= false
  break if done
  done = true
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1119,6 +1119,10 @@ condition](#safe-assignment-in-condition).
     end
     ```
 
+  :warning: Switching from e.g. `while` to `loop` changes the
+  scope of local variables. Variables defined in a `loop` go
+  out of scope after each iteration.
+
 * <a name="loop-with-break"></a>
   Use `Kernel#loop` with `break` rather than `begin/end/until` or
   `begin/end/while` for post-loop tests.


jaredbeck:
I'm "asking for a friend" :) 

(I made this mistake)

equivalent:
Good point but I'm not sure if it's Style Guide responsibility to point out API documentation differences

0x01f7:
Is it not a common sense that `if`/`while`/`for` will not introduce a new scope, but `Enumerate#each`/`Kernel#loop`/`any other method  with a block` will?

jaredbeck:
> Is it not .. common sense ..?

Not for me. `while true do` and `loop do` both have the `do` and `end` keywords, so to me they look very similar.

Swarzkopf314:
> Is it not a common sense (...)

Actually everything is a common sense if you're carefull enough. You're missing the point.

jaredbeck:
If an explicit warning just *feels like too much text*, we could instead *hint at* the difference:

```diff
- Use `Kernel#loop` instead of `while/until`
+ Pass a block to `Kernel#loop` instead of using the `while/until` keywords
```

I'm fine with either, though I expect an explicit warning to help more people.

dakull:
@jaredbeck :+1: whilst it's an API detail it's also a clearly misleading issue from a syntax POV

jaredbeck:
There is a precedent for warning people about tricky stuff, e.g. the suggestion to use `alias` instead of `alias_method` comes with this warning:

> Also be aware of how Ruby handles aliases and inheritance: an alias references the method that was resolved at the time the alias was defined; it is not dispatched dynamically.

0x01f7:
> Actually everything is a common sense if you're carefull enough.

Yes.

> It's up to Bozhidar, but there is a precedent for warning people about tricky stuff, ...

If you read the guide carefully, you may notice that there is a warning in rule [no-for-loops](https://github.com/bbatsov/ruby-style-guide#no-for-loops) which describes the differences. So don't put this stuff everywhere.

jaredbeck:
> If you read the guide carefully, you may notice that there is a warning in rule no-for-loops which describes the differences. So don't put this stuff everywhere.

I agree, let's keep it DRY. Both places should point to the same warning.

README.md
```diff
@@ -1119,6 +1119,10 @@ condition](#safe-assignment-in-condition).
     end
     ```
 
+  :warning: Switching from e.g. `while` to `loop` changes the
```
bbatsov: Probably this should be mentioned where we suggest `each` over `loop` instead.

bbatsov:
Guess we can close this one.

dakull:
Happy new year! 🎆 🌟 🦄

PR #541, opened 2016-02-12
Change Array check splat(*) example from "good" to "bad":

WizardOfOgz:
Using the splat technique `[*object]` will _always_ instantiate a new Array object, whereas
`Array(object)` will return the same object it is given as an argument if that argument is already an Array.

```ruby
> my_array = [1, 2, 3]
 => [1, 2, 3]

> my_array.object_id
 => 70240742761080

# This example is bad because it creates a new Array instance
# as demonstrated by showing that it has a different object_id.
>   [*my_array].object_id
 => 70240734757900

# This returns the same Array instance.
>   Array(my_array).object_id
 => 70240742761080
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1726,18 +1726,18 @@ no parameters.
   # => 'one, two, three'
   ```
 
-* <a name="splat-arrays"></a>
-  Use `[*var]` or `Array()` instead of explicit `Array` check, when dealing
+* <a name="array-coercion"></a>
+  Use `Array()` instead of explicit `Array` check or `[*var]`, when dealing
   with a variable you want to treat as an Array, but you're not certain it's an
   array.
-<sup>[[link](#splat-arrays)]</sup>
+<sup>[[link](#array-coercion)]</sup>
 
   ```Ruby
   # bad
   paths = [paths] unless paths.is_a? Array
   paths.each { |path| do_something(path) }
 
-  # good
+  # bad (always creates a new Array instance)
   [*paths].each { |path| do_something(path) }
 
   # good (and a bit more readable)


dakull:
and consequently an order of magnitude slower:

```ruby
require 'benchmark'

my_array = [1, 2, 3, 4]

N = 1000000

Benchmark.bmbm do |x|
  x.report("splat") { N.times { [*my_array] } }
  x.report("Array")  { N.times { Array(my_array) } }
end
```

```shell
            user     system      total        real
splat   0.250000   0.000000   0.250000 (  0.257150)
Array   0.070000   0.000000   0.070000 (  0.071887)
```

besen:
The change from 'good' to 'bad' doesn't agree with the description any more:

> Use `[*var]` or `Array()` instead of explicit `Array` [...]

You probably should update that too.

WizardOfOgz:
@besen Thanks, I have updated the updated the description

0x01f7:
What's purpose when people use this feature? Does anyone care about this difference(*it will be an new instance or not*)? I don't think so.

As it will casuse a performance issue, I prefer use `okish` instead of `bad`.

WizardOfOgz:
@0x01f7
> Does anyone care about this difference(it will be a new instance or not)? I don't think so.

Yes, people care greatly about this. It can have a significant negative impact on performance
if the code is called frequently. The difference may be negligible when your code runs once, but
when it runs millions of times in succession (a very realistic scenario for many of the things 
I work on) it may take a noticeably greater amount of time and memory (potentially triggering garbage collection).

> As it will casuse a performance issue, I prefer use okish instead of good.

A performance issue is _always_ bad, especially in this case since it is unnecessary. There is nothing "okish" about it. Duplicating objects in memory without need is very poor programming.
We should care enough about the code we write to make it as good as possible. It will make your applications better, and your employer 
and fellow developers will appreciate high code quality when you produce it.

bbatsov:
I'm fine with this change. Using `Array()` is obviously more readable as well.

PR #544, opened 2016-02-18
Update README.md

danlo:
I think you meant to put a "bad" here.


0x01f7:
Good point. But we should put `good` there, rather than `bad`.

bbatsov:
Yep, that's "good".

PR #546, opened 2016-02-22
add no greater-than rule

kholbekj:
This change, while maybe a bit controversial, is based on the ideas presented here:
http://llewellynfalco.blogspot.de/2016/02/dont-use-greater-than-sign-in.html

I have yet to hear drawbacks of this approach, and it greatly helps readability, especially for dates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,9 +293,9 @@ Translations of the guide are available in the following languages:
   case
     when song.name == 'Misty'
       puts 'Not again!'
-    when song.duration > 120
+    when 120 < song.duration
       puts 'Too long!'
-    when Time.now.hour > 21
+    when 21 < Time.now.hour
       puts "It's too late"
     else
       song.play
@@ -305,9 +305,9 @@ Translations of the guide are available in the following languages:
   case
   when song.name == 'Misty'
     puts 'Not again!'
-  when song.duration > 120
+  when 120 < song.duration
     puts 'Too long!'
-  when Time.now.hour > 21
+  when 21 < Time.now.hour
     puts "It's too late"
   else
     song.play
@@ -1036,12 +1036,12 @@ Translations of the guide are available in the following languages:
 
   ```Ruby
   # bad
-  if (x > 10)
+  if (10 < x)
     # body omitted
   end
 
   # good
-  if x > 10
+  if 10 < x
     # body omitted
   end
   ```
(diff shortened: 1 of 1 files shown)


dipth:
While I personally applaud the idea of clean and bug-free code, I also think that the rule of not allowing the use of `>` doesn't really result in that as it requires you to think very non-intuitively about a statement.

Ruby is a very expressive language that often prides itself about it's readability and flow in how it can be interpreted, but this rule hampers that expressiveness.

Take the following statement as an example:

`if user.age > 18`, here it is very clear that we are looking for users that are older than 18 years.
The human interpretation of that line would be something like this:
"If the age of the user is greater than 18"

Now take the counter-example:

`if 18 < user.age`, here you inverse the subject of the interrogation, instead of using the user as the subject we turn the age into the subject resulting in very angular and weird readability. The human interpretation would be something like: "If 18 is less than the users age"
As you can see there's a detachment from the user and we suddenly start treating the age as the subject.

In and by itself this example might seem like a minor issues, but imagine that you have a method with multiple statements mixed together where you have to switch out your idea of what the subject of a given statement is on every other line, then it could quickly lead to a lot of unnecessary mental overhead.

That is my opinion at least :)

kholbekj:
I disagree. I think confusion more often stems from assertions about their intend, which this is a perfect example of. 

`if user.age > 18` is something that you expect to bigger than 18 in a check, then you'd often miss it if it's the opposite way. In turn, I think the "time-line" structure of conditionals makes a lot of sense, as it always provides a visual clue as to which is supposed to be greater.

dipth:
> In turn, I think the "time-line" structure of conditionals makes a lot of sense, as it always provides a visual clue as to which is supposed to be greater.

But having the mentally visualise these statements as graphical timelines makes for the exact mental overhead that I am talking about. It adds confidence at the cost of intuitiveness and speed of interpretation.

dipth:
I also want to point out that you could also use Ruby's Comparable mixin to get the best of both worlds when comparing ranges:

```ruby
user.age.between? 18, 30
```

Ch4s3:
I completely agree with @dipth. `if foo.value > bar` is a clear and useful idiom if your are concerned with `foo` having a value greater than `bar`. I don't see how a number line adds clarity here. IMO semantics matter for comparisons, `bar < foo.value` si a context switch if the subject of the method is `foo`.

onebree:
I do not think this rule is relevant, as it does not apply only to Ruby. This is a _Ruby_ style guide, where the proposed rule can apply to any language with comparison operators.

I agree with @dipth, in that the `Comparable` lib should be used when possible. It reads more like English, which is a "feature" of Ruby in some ways.

PR #548, opened 2016-03-03
best practice to keep the expensive expression in the right of conditionals

shivabhusal:
- <a name="keep-expensive-operations-in-right-of-operators"></a>  In ruby conditionals the operand in left of `&&` operator is evaluated first and if 
  it is `false` the operand in the  right is not evaluated and saves execution time. 
  Similarly if left operand around `||` operator is evaluated to be `true` then the 
  operand at right side is skipped. So best practice is to keep the expensive expressions in the right side.
  <sup>[[link](#keep-expensive-operations-in-right-of-operators)]</sup>
  
  ``` ruby
  # bad
  if some_expensive_expression && simple_expression
    do_something
  end
  
  # good
  if simple_expression && some_expensive_expression
    do_something
  end
  
  # bad
  if some_expensive_expression || simple_expression
    do_something
  end
  
  # good
  if simple_expression || some_expensive_expression
    do_something
  end
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -960,6 +960,37 @@ Translations of the guide are available in the following languages:
   some_condition && do_something
   ```
 
+* <a name="keep-expensive-operations-in-right-of-operators"></a>
+  In ruby conditionals the operand in left of `&&` operator is evaluated first and if 
+  it is `false` the operand in the  right is not evaluated and saves execution time. 
+  Similarly if left operand around `||` operator is evaluated to be `true` then the 
+  operand at right side is skipped. So best 
+  practice is to keep the expensive expression in the right.
+<sup>[[link](#keep-expensive-operations-in-right-of-operators)]</sup>
+```ruby
+# bad
+if some_expensive_expression && simple_expression
+  do_something
+end
+
+# good
+if simple_expression && some_expensive_expression
+  do_something
+end
+
+
+# bad
+if some_expensive_expression || simple_expression
+  do_something
+end
+
+# good
+if simple_expression || some_expensive_expression
+  do_something
+end
+```
+
+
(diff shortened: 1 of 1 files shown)


jdickey:
Good point. *Better* practice, I'd argue, is to set ordering based on dependency; if that results in the expensive operation being performed first, then that's a code smell telling you that you probably want to look at refactoring. Cache the things you can, of course, but if the expensive operation depends on transient data, rethink and refactor.

Then again, it's not too difficult for developer time to outweigh the value of such refactoring. As a wet-behind-the-ears hacker with a mere half-dozen years in the craft, I once spent two weeks shaving half the time off an operation that, though performed daily and resource-intensive, delivered a cost savings over production life of approximately a quarter of the value of my time invested working on it.

That, two years later, won me a Faux Golden Goose Egg from the company; an annual recognition of well-intentioned, often in its way brilliant work that *just happened* to be a waste of company assets.

bipashant:
Good point :+1:

ne-ha:
:+1:

gkunwar:
Good point. :+1:

Iccr:
:+1:

dakull:
:+1: neat

equivalent:
:+1:

shivabhusal:
hi @bbatsov how about letting these changes to master? What do you say :)

ojab:
It's best practice, but it's about performance and has nothing to do with `style` this guide about.

bbatsov:
> It's best practice, but it's about performance and has nothing to do with style this guide about.

Indeed. Plus this advice is hardly Ruby-specific.

PR #552, opened 2016-03-27
Fix indentation in naming section and typo in error subclass section

akowalz:
Indentation was slightly off in naming section: 
https://github.com/bbatsov/ruby-style-guide#snake-case-symbols-methods-vars

And there was a typo in the section about error subclasses:
https://github.com/bbatsov/ruby-style-guide#standard-exceptions
> Favor the use of exceptions ~~for~~ from the standard library over introducing new exception classes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1990,7 +1990,7 @@ no parameters.
   end
 
   def SomeMethod
-   # some code
+    # some code
   end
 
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: Thanks!

PR #554, opened 2016-04-02
Omit parentheses additional condition

dsounded:
I've opened it, because many times I saw `defined?` vs `defined?()` and debates about "keyword" status of `puts`, but I do agree we need omit them though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1142,7 +1142,8 @@ condition](#safe-assignment-in-condition).
 * <a name="no-dsl-parens"></a>
   Omit parentheses around parameters for methods that are part of an internal
   DSL (e.g. Rake, Rails, RSpec), methods that have "keyword" status in Ruby
-  (e.g. `attr_reader`, `puts`) and attribute access methods. Use parentheses
+  (e.g. `attr_reader`, `attr_writer`, `alias_method`) and methods that have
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1142,7 +1142,8 @@ condition](#safe-assignment-in-condition).
 * <a name="no-dsl-parens"></a>
   Omit parentheses around parameters for methods that are part of an internal
   DSL (e.g. Rake, Rails, RSpec), methods that have "keyword" status in Ruby
-  (e.g. `attr_reader`, `puts`) and attribute access methods. Use parentheses
+  (e.g. `attr_reader`, `attr_writer`, `alias_method`) and methods that have
+  implicit receiver on a main object level (e.g. `puts`, `defined?`). Use parentheses
```
esdoppio: Isn't `defined?` actually a keyword?

bbatsov: It is.

DanRathbun: It is an [interpreter **operator**](http://ruby-doc.com/docs/ProgrammingRuby/html/tut_expressions.html#UG) only with the question mark and therefore "defined" is not a reserved word nor a keyword. (It would be valid to used "defined" as a local variable name or a method name, however perhaps not so smart to do so.)
Anyway it's rules perhaps should follow that of operators.

PR #560, opened 2016-04-21
Gather rules about spaces around braces in one place

esdoppio:
* https://github.com/bbatsov/ruby-style-guide#pad-string-interpolation

  This rule is really about code layout instead of strings. Makes more sense to put it in the "Source Code Layout" section.

* https://github.com/bbatsov/ruby-style-guide#spaces-operators

  By splitting out the `{ }` part and putting it together with `( )` and `[ ]`, it would make the rules in question easier to follow, compare and locate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,6 @@ Translations of the guide are available in the following languages:
   ```Ruby
   sum = 1 + 2
   a, b = 1, 2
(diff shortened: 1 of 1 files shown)


bbatsov:
👍 Fine by me. Thanks!

PR #565, opened 2016-04-25
Hyphenate 'multi-line' consistently

landongrindheim:
Not meant to be a nitpick. I stumbled upon this when using ctrl+f to quickly lookup rules for multi-line blocks.


bbatsov:
👍

PR #569, opened 2016-05-15
Revert "Remove duplicate section "pad-string-interpolation""

maiamatheus:
Reverts bbatsov/ruby-style-guide#568


esdoppio:
@maiamatheus the reason of reverting?

PR #570, opened 2016-05-18
Gather rules of the use of parentheses in method invocations in one entry

esdoppio:
The rules are scattered around. Gathered and reorganized them for easy reference.

@bbatsov do u think this is too much for a single entry? Maybe I can split it in two?


bbatsov:
I think this looks OK. Thanks!

PR #575, opened 2016-05-27
proposed memoizing style

ivar:
I've seen a few variable initialization styles and this is what we ended up standardizing on for my current team. I'm curious as to how the community at large feels about this particular approach. 

Regardless of what ends up being considered the correct style for perform initialization, it'd be good to get something like this in the Mr. Batsov's styleguide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1446,6 +1446,26 @@ condition](#safe-assignment-in-condition).
   name ||= 'Bozhidar'
   ```
 
+* <a name="no-double-pipe-for-block"></a>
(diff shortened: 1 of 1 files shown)


0x01f7:
You can't use `return` keyword or any instance variables unless in a method declaration. And can you explain why the first style is bad. Thanks.

arr-dev:
I believe the accepted style for multi-line is:

```rb
@name ||= begin
  # compute
end
```

WaKeMaTTa:
```ruby
return @name unless name.nil
```

bbatsov:
@arr-dev is right.

PR #577, opened 2016-06-02
[Add] Vietnamese version

CQBinh:
90% translated.


bbatsov:
I'm guessing the old version is less than 90% translated, right?

CQBinh:
The old version is about 10->15% translated.
Only the introduction is translated (`Prelude` and `The Ruby Style Guide`) sections.
https://github.com/scrum2b/ruby-style-guide/blob/master/README-viVN.md

bbatsov:
👍

PR #578, opened 2016-06-02
Add section for octal literals

tejasbubane:
Prefer `0o` over `0` and `0O` for octal literals.

Related to https://github.com/bbatsov/rubocop/pull/3177

PR #579, opened 2016-06-02
Add rules for hexadecimal, binary and decimal literal prefixes

tejasbubane:
Related to #578

@bbatsov suggested to add rule for hex numbers. I went ahead and added
one for binary as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -569,19 +569,26 @@ Translations of the guide are available in the following languages:
   num = 1_000_000
   ```
 
-* <a name="octal-literals"></a>
-  Use `0o` for octal literals instead of `0` or `0O`.
-<sup>[[link](#octal-literals)]</sup>
+* <a name="numeric-literal-prefixes"></a>
+  Prefer smallcase letters for numeric literal prefixes.
+  `0o` for octal, `0x` for hexadecimal and `0b` for binary.
+  Do not use `0d` prefix for decimal literals.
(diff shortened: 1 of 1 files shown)


cremno:
There's also the rather pointless `0d`/`0D` prefix. I also think this can be condensed into one rule. Having 3 or even 4 very similar sections is kinda ugly.

tejasbubane:
@cremno Thanks! Did not know about the `0d` prefix. I have added it but since it is pointless, not sure if we should keep it. 

And yes, the 4 very similar sections looked kinda ugly. I have combined them into a single section.

cremno:
I think you've misunderstood. The following would make sense to me:

- use `0[bodx]` instead of `0[BODX]`
- use `0o` instead of `0` (like now but minus `0O` - it doesn't matter anyway if it's `00` or `0O`)

And maybe (@bbatsov?):

- don't use `0d`

tejasbubane:
`0[bodx]` looks a little confusing to me.

cremno:
My comment is just an example of how I would arrange the sections. I didn't want to propose anything in it as official wording or criticize yours.

tejasbubane:
@cremno Yes, I understand. I just felt it a little confusing. Let's see what @bbatsov has to say.

bbatsov:
Looks good to me.

PR #582, opened 2016-06-03
Add accessor and mutator method names conventions

tejasbubane:
Add a section to avoid using accessor and mutator method names starting
with `get_` and `set_`.

This came up in https://github.com/bbatsov/rubocop/issues/3181

### DIFF
--- a/README.md
+++ b/README.md
@@ -2539,6 +2539,37 @@ no parameters.
   end
   ```
 
+* <a name="accessor_mutator_method_names"></a>
+  For accessors and mutators, avoid prefixing method names with
+  `get_` and `set_`.
+  It is a Ruby convention to use attribute names for accessors (readers) and
+  `attr_name=` for mutators (writers).
+<sup>[[link](#accessor_mutator_method_names)]</sup>
+
+  ```Ruby
+  # bad
+  class Person
+    def get_name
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2539,6 +2539,37 @@ no parameters.
   end
   ```
 
+* <a name="accessor_mutator_method_names"></a>
+  For more complex accessors and mutators, avoid prefixing methods with
```
bbatsov: remove "more complex" here. I understand what you mean, but I believe this wording is more confusing than helpful.
methods -> method names

README.md
```diff
@@ -2539,6 +2539,37 @@ no parameters.
...
+  `get_` and `set_`.
+  It is a ruby convention to use direct names for accessors (readers) and
```
bbatsov: ruby -> Ruby

bbatsov: direct names -> attribute names (or something like this)

Swarzkopf314: However its worth mentioning that overriding attribute accessors in Rails
can easily backfire at you. In sich cases I prefer to use get_/set_.
5 cze 2016 09:08 "Bozhidar Batsov" <notifications@github.com> napisał(a):

tejasbubane:
@bbatsov fixed as per your comments.

PR #587, opened 2016-06-30
Add information about Time versus DateTime

Drenmi:
As per the [documentation](http://ruby-doc.org/stdlib-2.3.0/libdoc/date/rdoc/DateTime.html), `DateTime` should only be used in cases where historical calendar reform needs to be taken into consideration. In all other cases, using `Time` or `Date` is advisable.

This clause will also support a Rubocop check being planned in https://github.com/bbatsov/rubocop/issues/1952.


bbatsov:
👍

PR #589, opened 2016-07-10
Improve a number of punctuation/minor grammatical issues

ChallahuAkbar:
I noticed that there were quite a few instances where hyphens were used in place of dashes; I also noticed that there were a number of potentially confusing instances where a single code block is used to talk about "`if` or `unless`" (e.g., "use `if/unless` instead"). I changed this to place the slash between two code blocks, to differentiate from the instances of talking about an `if/then/else/end` construction or the like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,9 @@ By the way, if you're into Rails you might want to check out the complementary
 
 This Ruby style guide recommends best practices so that real-world Ruby
 programmers can write code that can be maintained by other real-world Ruby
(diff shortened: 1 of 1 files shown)


bbatsov:
The main reason I don't typically use dashes is that I can't quickly type them. :-) Thanks!

ChallahuAkbar:
Ah yeah makes sense, in Markdown I just used the HTML `&mdash`. Also, I tend to run Windows on my computer w/ Linux in a VM, so I just ALT+0151 for the em dash.

PR #593, opened 2016-08-03
fix misplaced parentheses in parentheses omitting section

hedgesky:
In this section we are going to advice omitting parentheses, but doing the opposite instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,9 +717,9 @@ Translations of the guide are available in the following languages:
 
     ```Ruby
     # bad
-    expect(bowling.score).to eq 0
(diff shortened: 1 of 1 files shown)


hedgesky:
I've updated this PR. Now simpler example which we discussed in #611 is used.

bbatsov:
I guess we could have used something even simpler - like `attr_accessor`, but this will do.

hedgesky:
`attr_accessor` is reserved for the next section: "Methods that have "keyword" status in Ruby" 😄 . Thanks for merging this!

PR #595, opened 2016-08-18
`has_key?` description note is giving false impression

mentero:
Hi

Currently in a section regarding `has_key?` vs `key?` there is a sentence linking to an old email which was used as a rationale for this rule. It is also noted that the `has_key?` is considered deprecated. 

I believe that this note is unnecessary and can give a false impression. `has_key?` is not considered deprecated and is still actively in use. Matz clearly stated that it is not going anywhere here https://github.com/bbatsov/rubocop/issues/3224 

This makes me believe that the note about `has_key?` deprecation in Matz's mail was simply an unfortunately phrased opinion on a personal preference and is not an official statement of a core team. 

To back it up this rule started a discussion in my team today and this part of style guide was brought up as an argument while I firmly believe that choosing between `has_key?` and `key?` is highly subjective matter and one is arguably better than the other. This is especially true since the rationale behind this rule is no longer valid.

---

Looks like my editor also fixed some trailing whitespaces ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -995,7 +995,7 @@ Translations of the guide are available in the following languages:
   Avoid the use of `!!`.
 <sup>[[link](#no-bang-bang)]</sup>
 
-  `!!` converts a value to boolean, but you don't need this explicit 
+  `!!` converts a value to boolean, but you don't need this explicit
   conversion in the condition of a control expression; using it only
   obscures your intention. If you want to do a `nil` check, use `nil?`
   instead.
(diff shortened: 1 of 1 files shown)


booch:
If they're no longer (and maybe never have been) deprecated, it would seem that the *longer* versions should be preferred, as they're more clearly intention-revealing.

bbatsov:
> If they're no longer (and maybe never have been) deprecated, it would seem that the longer versions should be preferred, as they're more clearly intention-revealing.

They are still inconsistent with most predicates in the core and standard library.

PR #596, opened 2016-08-23
`unless` to `if`

dansteele:
I believe that this makes more sense, and was my immediate feeling when seeing it.

I view unless as:

> I'd normally do this thing, `unless` this condition

In this case, you're not 'normally' calling next.

Let me know your thoughts!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1929,7 +1929,7 @@ no parameters.
 
   # good
   [0, 1, 2, 3].each do |item|
-    next unless item > 1
+    next if item < 1
     puts item
   end
   ```


booch:
Your change is not equivalent to the "bad" example. You'd need `next if item <= 1`.

To the general point, I don't think there's a terribly strong argument either way. It depends upon how you're thinking about the condition. Honestly, I think it's a bad example with respect to `if`/`unless`, because it's too easy to change the direction of the `<`/`>`.

I think it'd also be nice to point out that this is basically a guard clause inside a loop. Maybe there's some mention of guard clauses elsewhere that has a better example we could borrow.

bbatsov:
I agree with @booch. Guess we can close this.

PR #597, opened 2016-09-12
add notion about default rake task

hedgesky:
I've seen this as a rather widely used approach. For example, it's used for running tests in Rails (both [itself](https://github.com/rails/rails/blob/master/Rakefile#L17) and Rails apps), [Sinatra](https://github.com/sinatra/sinatra/blob/master/Rakefile#L6), [rubocop](https://github.com/bbatsov/rubocop/blob/master/Rakefile#L35), [thin](https://github.com/macournoyer/thin/blob/master/Rakefile#L8), [hanami](https://github.com/hanami/hanami/blob/master/Rakefile#L17), etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4006,6 +4006,18 @@ resource cleanup when possible.
   Avoid more than three levels of block nesting.
 <sup>[[link](#three-is-the-number-thou-shalt-count)]</sup>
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Seems to me this might be outside the scope of the style guide.

PR #600, opened 2016-09-23
Proposing rule for referencing class methods in other class methods.

jdickey:
Using the example of a "service object" and its standard `.call` class
method that instantiates the class it's a member of, this proposer
suggests that class methods simply call other class methods, rather than
regurgitating the class name or even the `self` keyword. Many people,
including this proposer, tend to forget that `new` will (generally) Do
The Right Thing in such situations without decoration _because_ the
method in `BasicObject` is overridden appropriately in `Class` and
elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2838,6 +2838,35 @@ no parameters.
   end
   ```
 
+* <a name="class-and-self"></a>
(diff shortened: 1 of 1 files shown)


bbatsov:
Sounds good to me.

PR #605, opened 2016-10-15
Advise to avoid guard clauses on multiline statements

tessi:
This  is proposed according to the discussion in https://github.com/bbatsov/rubocop/issues/3570

The reasoning was that guard clauses get confusing when they are too far away from the beginning of their statement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1933,6 +1933,34 @@ no parameters.
   end
   ```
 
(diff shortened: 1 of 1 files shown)


tessi:
call me blind. it's already in there. https://github.com/bbatsov/ruby-style-guide#no-multiline-if-modifiers

PR #608, opened 2016-11-03
Add rule to avoid chaining methods after `end`

gnapse:


equivalent:
I disagree with this rule, the reason is that some developers prefer this and there is no harm *in case they follow other indentation rules*  eg:

```ruby
[1,2,4]
  .map do |x|
    #something that require multiple lines
    (x * 400 / 15) +200
  end
  .select { |x| x>10 }
  .map{ |x| x.to_s}
```

as long as it's on separate line, so that source control versioning (e.g. git) pick up the change then it's ok

in your example

````
  if target.is_a?(Role)
   target
  else
    target.role
  end.destroy
````

I would argue to do:
````
  if target.is_a?(Role)
   target
  else
    target.role
  end
    .destroy
````

**Comment by gnapse** [comment]
I added this as an extension of the rule just above it ([link](https://github.com/bbatsov/ruby-style-guide#single-line-blocks)). In it, it is said that one should "avoid `do..end` when chaining". And I personally agree to the point where I think most chaining of the form `end.something_else` is bad practice.

Although I agree with you that your first example is ok, I had not thought of that. But the one with the `if` is not, IMO. I would always suggest in such cases to store the result of the `if` expression, and then use that for whatever else is needed.

Anyway, I would amend the rule to accommodate for your first example, but I guess this is still a somewhat controversial topic that might not get consensus. I just wanted to brought up the issue to see how it's received. Thanks for your input.

**Comment by mentero** [comment]
I don't have a strong opinion about chaining after `end`, but can we provide examples without the use of `is_a?`? I think it will be quite easy to come by with such example and we won't suggest that using `is_a?` is a good practice because it isn't. It violates design rules such as _Tell don't Ask_ and should be replaced by _polymorphism_ unless completely necessary.

**Comment by bbatsov** [comment]
I'm fine with the addition of this rule, although there should probably be more clarity about the types of blocks. Probably the preceding rule could be broken into two and extended a bit. I don't like it's current form that much.

**Comment by pirj** [comment]
This is pretty much covered in https://github.com/rubocop-hq/ruby-style-guide/blob/master/README.adoc#single-line-blocks-delimiters:

>  Avoid `do…​end` when chaining.

```ruby
# bad
names.select do |name|
  name.start_with?('S')
end.map { |name| name.upcase }
```

Despite some downvotes, I believe it is a reasonable guideline. Thank you!

Issue #606, opened 2016-10-21
No rule for snake_case with number


jessiel-hacke:
There is no rule for a variable name with numbers.

For example if I have a variable named like `a_400`, should it be `a400` instead?

Having

```
# bad
a_400 = something

# good
a400 = something
```

or both situations are good?

RomanADavis:
Wouldn't it be better just to avoid numbers in variable names? If you need a1 and an a2, wouldn't you rather make an array a with a[1] and an a[2]?

Am I missing a use case here?

RomanADavis:
Wouldn't it be better just to avoid numbers in variable names? If you need a1 and an a2, wouldn't you rather make an array a with a[1] and an a[2]?

Am I missing a use case here?

RomanADavis:
Wouldn't it be better just to avoid numbers in variable names? If you need a1 and an a2, wouldn't you rather make an array a with a[1] and an a[2]?

Am I missing a use case here?

RomanADavis:
Wouldn't it be better just to avoid numbers in variable names? If you need a1 and an a2, wouldn't you rather make an array a with a[1] and an a[2]?

Am I missing a use case here?

RomanADavis:
Wouldn't it be better just to avoid numbers in variable names? If you need a1 and an a2, wouldn't you rather make an array a with a[1] and an a[2]?

Am I missing a use case here?

RomanADavis:
Wouldn't it be better just to avoid numbers in variable names? If you need a1 and an a2, wouldn't you rather make an array a with a[1] and an a[2]?

Am I missing a use case here?

RomanADavis:
Wouldn't it be better just to avoid numbers in variable names? If you need a1 and an a2, wouldn't you rather make an array a with a[1] and an a[2]?

Am I missing a use case here?

jessiel-hacke:
I think there are a few use cases where I could use this format of variable naming. As an example, cnab400, which is a format for bank files used on batch payments. Maybe I should use a different name for the variable but it would be nice to have it on the style guide saying that I should use a different format.

eliotsykes:
**TLDR** In the open source Ruby projects searched, not using underscores in numbered variables happens more frequently (e.g. `a1` is used more often than `a_1`).

---

Without opinion, in case it helps, here's some (probably flawed) stats about usage of `_1` vs `1` suffixes, taken from [searching 100+ ruby codebases](/eliotsykes/real-world-rails):

```
[...]

PR #609, opened 2016-11-11
Add rule when variable name has a number

jessiel-hacke:


bbatsov:
While I don't mind adding a rule for this, the decision which is the better style sounds arbitrary. Have done any analysis of what's more common in general and why?

strika:
There is a discussion and some analysis on the related issue - https://github.com/bbatsov/ruby-style-guide/issues/606.

bbatsov:
👍

Issue #76, opened 2012-02-22
Trailing Comma in Multi-Line Hashes & Arrays?


benhamill:
I think I first ran into this style in Python, but I can't be sure. The idea is that in a multi-line liter Hash or Array, you leave a trailing comma on the final element, thus:

``` ruby
DEFAULTS = {
  page: 1,
  per_page: 50,
  sort_by: name,
}
```

The thinking is that it makes adding elements easier; it's easy to forget to add a trailing comma to the line before your new element, you don't clutter a diff by changing a line that, really, didn't have to do with your changes and you can copy/paste any line up or down the list without having to edit it.

Personally, I like this style, but I know others disagree. I thought I'd see what people thought here. So... thoughts?

jtara:
FWIW, recent versions of Javascript accept this, as well.

While I like the neatness of leaving the trailing comma off, it's bit me more times than I can think (when re-ordering items), and really should train myself into always adding the trailing comma.

Here's a use case that might be non-obvious, though really doesn't apply to a style-guide: generated code:

I treat all my Ruby files (and all other text files, for that matter...) as {{mustache}} templates in a pre-build step, so that product configuration can be incorporated from some YML configuration files. (Typically configuring multiple products from the same codebase.)

So, I sometimes do something like this:

``` ruby
    SOFT_ERRORS = [ {{#api.soft_errors}} '{{.}}', {{/api.soft_errors}} ]
```

This gives me e.g.

``` ruby
    SOFT_ERRORS = [  '12345',  '54321',  '22222',  ]
```

bk322:
This is surprising that ruby doesn't allow things like
```ruby
DEFAULTS = {
  page: 1
, per_page: 50
, sort_by: name
}
```
given it focus on syntax sugar. I wonder if there is a special reason for that.

Crenshinibon:
I recently did a lot of CoffeeScript and wondered why this isn't working:

```
DEFAULTS = {
  page: 1
  per_page: 50
  sort_by: name
}
```

varcharlez:
Wow, surprised that people are actually against this. Having a trailing comma is good practice as 1) it saves errors when future maintainers add an item to the hash, and 2) it preserves diffs so that when new items are added, they are the only thing that comes up in the diff.

franzliedke:
Indeed, this is very common in some other languages I know.

[...]

PR #610, opened 2016-11-14
comma for last element in multi line Array or Hash is valid - from point of source control 

equivalent:
The reasoning for this was originally that "because it's easier to add elements" which I agree is not good point on it's own -> [discussion](https://github.com/bbatsov/ruby-style-guide/issues/76).

But there is far bigger point missed from perspective of version source control (like in Git) in a sense that one line change requires another line to be marked as changed in source control

So if you have 

```
{
  foo: 'foo'
}
```

....and you introduce `bar: :bar`

```
{
  foo: 'foo',
  bar: :bar
}
```

the source control will mark both lines as changed as you introduced a comma

```git
    {
+     foo: 'foo',
+     bar: :bar
    }
```

Where as you have that line with comma from beginning:

```
{
  foo: 'foo',
}
```

... and you introduce new element on  a new line

```
{
  foo: 'foo',
  bar: :bar,
}
```

source control will show that only one line has changed:

```git
    {
      foo: 'foo',
+     bar: :bar,
    }
```

And this is far bigger argument than cleanliness

### DIFF
--- a/README.md
+++ b/README.md
@@ -3134,12 +3134,23 @@ resource cleanup when possible.
   ```
 
 * <a name="no-trailing-array-commas"></a>
-  Avoid comma after the last item of an `Array` or `Hash` literal, especially
-  when the items are not on separate lines.
+  Use of comma after the last item of an `Array` or `Hash` literal is 
+  acceptable when you are dealing with items on multiple lines, as
+  it prevents source controls versioning (e.g. Git) to point to multiple
+  lines when you introduce a change.
+  
+  Avoid comma after the last item when the items are not on separate lines.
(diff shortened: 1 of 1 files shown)


dre-hh:
and also favored by other communities as well 
https://github.com/airbnb/javascript#commas--dangling

pirj:
Arguments are strong enough from both sides. It indicates no clear consensus.

> There are some areas in which there is no clear consensus in the Ruby community regarding a particular style ... In such scenarios all popular styles are acknowledged and it’s up to you to pick one and apply it consistently.

Just for reference, [original discussion in a now-closed ticket](https://github.com/rubocop-hq/ruby-style-guide/issues/76#event-11262787).

PR #611, opened 2016-11-18
Fix misleading typo in the internal DSL parentheses omission rule

jibiel:
It's either that or the example should be moved out of the "Only omit parentheses for" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,9 +717,9 @@ Translations of the guide are available in the following languages:
 
     ```Ruby
     # bad
-    expect(bowling.score).to eq 0
-    # good
     expect(bowling.score).to eq(0)
+    # good
+    expect(bowling.score).to eq 0
     ```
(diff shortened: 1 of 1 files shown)


dsounded:
I would say here it should `be_zero` matcher

bbatsov:
I guess the example is pretty bad overall, as it actually features DSL methods - `to` and `eq`. Maybe we should come up with something simpler...

hedgesky:
Maybe we could use this example here?

```ruby
# bad 
validates(:name, presence: true)
# good
validates :name, presence: true
```

bbatsov:
Yeah, something like this would do. Obviously unambiguous examples are better. :-)

bbatsov:
Duplicate of https://github.com/bbatsov/ruby-style-guide/pull/593

PR #623, opened 2017-01-30
Do not use unless with multiple conditions

ProZsolt:


thepeoplesbourgeois:
Your example encourages developers to attempt to perform Boolean algebra on potentially long conditionals. Remembering to toggle the `!` on each condition, as well as to toggle between `&&` and `||` usages throughout the condition chain is a rather taxing mental task.

```ruby
# bad
unless definitely_bad? || !good? && good_for_nothing?
  do_something
end

# good?
if !definitely_bad? && good? || !good_for_nothing? # I don't even know if this all evaluates the same
  do_something
end
```

What do you think of changing the good example to `if !(conditions)`?

```ruby
# bad
unless definitely_bad? || !good? && good_for_nothing?
  do_something
end

# good
if !(definitely_bad? || !good? && good_for_nothing?)
  do_something
end
```

gnapse:
I wonder why this hasn't been discussed more since it was proposed, let alone why it has not been accepted. I too find it very odd to read complex `unless` conditions.

gnapse:
> Remembering to toggle the ! on each condition, as well as to toggle between && and || usages throughout the condition chain is a rather taxing mental task.

I'd rather have the author of the code do this once, than having the readers of the code having to decode such a complex boolean expression every time they read the code.

And remember that the author of the code is also a future reader of it.

mikegee:
Instead of the boolean algebra, the other solution to "unless with multiple conditions" is extracting the conditions to a method.

gnapse:
Yes, that would be a solution too. Anything that avoids a final result where you have something like

```ruby
unless something || sometingElse
```

That's up to the author of the code to refactor their code when it is flagged as violating this proposed rule.

thepeoplesbourgeois:
To each their own. I'm going to go ahead and say it's not ridiculous to say

```ruby
unless weekend? || on_vacation? || on_call?
  come_in_to_work
end
```

is more readable than

```ruby
if !weekend? && !on_vacation? && !on_call?
  come_in_to_work
end
```

jonathonjones:
Consider this case:

```ruby
unless admin? || superuser?
  return "You are not authorized"
end
```

That feels much more natural to me than

```ruby
if !admin? && !superuser?
  return "You are not authorized"
end
```

In English wouldn't you explain the requirements as "they must be either and admin or a super user to perform this action" rather than "they are not allowed to perform this action if they are not an admin and not a super user"?

PR #625, opened 2017-02-05
Typo s/when x;/when x:

amerov:


0x01f7:
It is not a typo. It should be interpreted as: Prefer `when x then ...` over `when x; ...` for one-line case.

PR #633, opened 2017-02-13
Update README.md

hachi8833:
On #snake-case-symbols-methods-vars-with-numbers, the bad sample `var_10  = 10` and good sample `var10    = 10` should be in the next #snake-case-symbols-methods-vars-with-numbers.

Thank you,


bbatsov:
Good catch!

We could have also just removed those, but that's not a big deal.

PR #636, opened 2017-02-19
Clarify a intention of a description for method-invocation-parens style

### LINKED ISSUE #635: Meaning of "Only omit parentheses for" is confusing in method-invocation-parens style
In method-invocation-parens style, The meaning of "Only omit parentheses for" is ambiguous.
https://github.com/bbatsov/ruby-style-guide/blob/8b6944d0179da48c5e71515dcf5baab4bd26ad34/README.md#method-invocation-parens

>   Only omit parentheses for
> 
>   * Method calls with no arguments:
> 
>     ```Ruby
>     # bad
>     Kernel.exit!()
>     2.even?()
>     fork()
>     'test'.upcase()
> 
>     # good
>     Kernel.exit!
>     2.even?
>     fork
>     'test'.upcase
>     ```
> 
>   * Methods that are part of an internal DSL (e.g., Rake, Rails, RSpec):
> 
>     ```Ruby
>     # bad
>     validates(:name, presence: true)
>     # good
>     validates :name, presence: true
>     ```
> 
>   * Methods that have "keyword" status in Ruby:
> 
>     ```Ruby
>     class Person
>       # bad
>       attr_reader(:name, :age)
>       # good
>       attr_reader :name, :age
> 
>       # body omitted
>     end
> 
>     # bad
>     puts(temperance.age)
>     # good
>     puts temperance.age
>     ```

I think the sentence has the following two meanings.

- **Must** only omit parentheses for
- only **can** omit parentheses for

And I think both of the two meanings are not appropriate.

First, in the case of "MUST omit", I think the `puts` example is not appropriate. Because I don't omit parentheses for `puts` sometime. 
This also applies to methods defined in `Kernel` other than` puts`(e.g. `Kernel.#sytem`).
For these methods, I think we may omit parentheses or not.
 https://ruby-doc.org/core-2.4.0/Kernel.html

So, "**MUST** omit" is not appropriate.

--------

Second, in the case of "can omit", this is not appropriate for other examples.
For example, I think parentheses for without arguments or DSL should be omitted. 

Improvement plan
---------

I propose an improvement plan.
I think we should divide into two sections, "Must omit" and "Can omit".

For example.

````markdown
  Must omit parentheses for

  * Method calls with no arguments:

    ```Ruby
    # bad
    Kernel.exit!()
    2.even?()
    fork()
    'test'.upcase()

    # good
    Kernel.exit!
    2.even?
    fork
    'test'.upcase
    ```

  * Methods that are part of an internal DSL (e.g., Rake, Rails, RSpec):

    ```Ruby
    # bad
    validates(:name, presence: true)
    # good
    validates :name, presence: true
    ```

  * Methods that have "keyword" status in Ruby:

    ```Ruby
    class Person
      # bad
      attr_reader(:name, :age)
      # good
      attr_reader :name, :age

      # body omitted
    end
    ```

  Can omit parentheses for

  * Methods that have "keyword" status in Ruby, but not declarative:

    ```Ruby
    # good
    puts(temperance.age)
    system('ls')
    # also good
    puts temperance.age
    system 'ls'
    ```
````

WDYT?

### PR DESCRIPTION
**Pull request by pocke**
Fix #635

### DIFF
--- a/README.md
+++ b/README.md
@@ -778,7 +778,7 @@ Translations of the guide are available in the following languages:
   temperance = Person.new('Temperance', 30)
   ```
 
-  Must omit parentheses for
+  Always omit parentheses for
 
   * Method calls with no arguments:
 
@@ -820,7 +820,7 @@ Translations of the guide are available in the following languages:
 
   Can omit parentheses for
 
-  * Methods that have "keyword" status in Ruby, but not declarative:
+  * Methods that have "keyword" status in Ruby, but are not declarative:
 
     ```Ruby
     # good
--- a/README.md
+++ b/README.md
@@ -778,7 +778,7 @@ Translations of the guide are available in the following languages:
   temperance = Person.new('Temperance', 30)
   ```
 
-  Only omit parentheses for
+  Must omit parentheses for
 
   * Method calls with no arguments:
 
@@ -816,13 +816,20 @@ Translations of the guide are available in the following languages:
 
       # body omitted
     end
-
-    # bad
-    puts(temperance.age)
-    # good
-    puts temperance.age
     ```
 
+  Can omit parentheses for
+
+  * Methods that have "keyword" status in Ruby, but not declarative:
+
+    ```Ruby
+    # good
+    puts(temperance.age)
+    system('ls')
+    # also good
+    puts temperance.age
+    system 'ls'
+    ```
 * <a name="optional-arguments"></a>
     Define optional arguments at the end of the list of arguments.
     Ruby has some unexpected results when calling methods that have

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -778,7 +778,7 @@ Translations of the guide are available in the following languages:
   temperance = Person.new('Temperance', 30)
   ```
 
-  Only omit parentheses for
+  Must omit parentheses for
```
bbatsov: Must -> Always

README.md
```diff
@@ -816,13 +816,20 @@ Translations of the guide are available in the following languages:
 
       # body omitted
     end
+    ```
 
-    # bad
-    puts(temperance.age)
+  Can omit parentheses for
+
+  * Methods that have "keyword" status in Ruby, but not declarative:
```
bbatsov: but are not

pocke:
@bbatsov Thanks for review. I've updated them.

pocke:
@bbatsov ping

Issue #538, opened 2016-01-30
Prefer instance var over attr_* methods inside class


adsteel:
It is easier to understand and slightly faster to directly access the instance variable inside the class.

``` ruby
# bad
class Person
  attr_reader :age

  def initialize(age)
    @age = age
  end

  def half_age
    age * 0.5
  end
end

# good
class Person
  attr_reader :age

  def initialize(age)
    @age = age
  end

  def half_age
    @age * 0.5
  end
end
```

The other side of that is, if the accessor is explicitly defined, that custom accessor should be preferred over directly using the instance variable.

``` ruby
# bad
class Person
  def initialize(age)
    @age = age
  end

  def age=(new_age)
    fail 'Can only get older' if @age > new_age

    @age = new_age
  end

  def celebrate_birthday
    @age = @age + 1
  end
end

# good
class Person
  def initialize(age)
    @age = age
  end

  def age=(new_age)
    fail 'Can only get older' if @age > new_age

    @age = new_age
  end

  def celebrate_birthday
    age = @age + 1
  end
end
```

jdickey:
I'd argue against this. If you start out using an instance variable and then later change to a custom accessor, you now must go and replace every use of the ivar with the accessor. Risk of breakage increases as a power of the number of changes made, so this could simply inhibit needed refactoring later.

adsteel:
@jdickey That's a really good argument against.

equivalent:
+1 on jdickey's argument

`named variables` are essential in order for better re-factoring. 

```ruby
class Foo
  def call
    user_ids = [1,2,3,4]
    user_ids.each { |id| puts id }
  end
end

# vs 

class Bar
  def call
    @user_ids = [1,2,3,4]
    @user_ids.each { |id| puts id }
  end
end
```

to 

```ruby
class Foo # or Bar
  def call
    user_ids.each { |id| puts id }
  end

  def user_ids
    [1,2,3,4]
  end
end
```

adsteel:
@jdickey @equivalent Would it be worth saying that all instance variables should be accessed with accessor methods? This would make for a consistent style, as well as easily allow for custom accessors if needed.

equivalent:
@adsteel  well yes and no. it really depends what do you want to achieve.

I honestly never touch instance variable in a class I always use attr_reader.

```ruby
class Foo
  attr_reader :foo

  def initialize(foo)
    @foo = foo
  end

  def do_stuff
    foo + 'bar'
[...]

PR #641, opened 2017-03-02
Consistent use of accessor in examples

hughevans:
The style guide implies that inside your class you should use the accessor methods `attr_reader` creates and not the ivars directly. Issue #538 questioned if this should be the other way around, but it appears not.

This PR changes the `def to_s` example to not use the ivar directly in accordance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2705,7 +2705,7 @@ no parameters.
     end
 
     def to_s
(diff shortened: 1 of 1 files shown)


bbatsov:
Fine by me. Thanks!

PR #644, opened 2017-03-14
clarify what is good and bad - method naming

BKSpurgeon:
hi all, please vet thoroughly before accepting. my understanding may well be wrong. chrs!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2104,13 +2104,16 @@ no parameters.
   :SomeSymbol
   :someSymbol
 
+ # bad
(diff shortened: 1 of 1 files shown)


andyw8:
Hi @BKSpurgeon, thanks for your contribution.

I think it's reasonably clear that anything below `# good` is consider 'good' until the `# bad` separator.

As this a very old PR, I'm going to close it, but please feel free to re-open the discussion if you think we should reconsider.

PR #646, opened 2017-03-16
Change "Camel Case" to "Pascal Case"

jimmcnulty41:
CamelCase is ambiguous as to whether the first letter should be lower or uppercase. PascalCase always means the first letter it capitalized, which we are recommending here

### DIFF
--- a/README.md
+++ b/README.md
@@ -2152,7 +2152,7 @@ no parameters.
 
 
 * <a name="pascalcase-classes"></a>
-  Use `PascalCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
+  Use `UpperCamelCase` (PascalCase) for classes and modules.  (Keep acronyms like HTTP, RFC, XML
   uppercase.)
 <sup>[[link](#pascalcase-classes)]</sup>
 
--- a/README.md
+++ b/README.md
@@ -2151,10 +2151,10 @@ no parameters.
   ```
 
 
-* <a name="camelcase-classes"></a>
-  Use `CamelCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
+* <a name="pascalcase-classes"></a>
+  Use `PascalCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
   uppercase.)
-<sup>[[link](#camelcase-classes)]</sup>
+<sup>[[link](#pascalcase-classes)]</sup>
 
   ```Ruby
   # bad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2151,10 +2151,10 @@ no parameters.
   ```
 
 
-* <a name="camelcase-classes"></a>
-  Use `CamelCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
+* <a name="pascalcase-classes"></a>
```
pocke: I guess the anchor should not change immoderately. So, I would like to still use `camelcase-classes` as an anchor, and use `Upper Camel Case` instead of `Pascal Case`. WDYT?

booch:
I don't think "PascalCase" is nearly as well-understood as "CamelCase". Plus it refers to another language that people might not be familiar with.

If the concern is not being clear that these should be uppercase CamelCase, then we should just add that clarification. Although classes and module names will have to be uppercase anyway, due to the fact that they're constants. (Assuming that you're using them in the normal way.)

bbatsov:
camelCase vs PascalCase :-)

README.md
```diff
@@ -2151,10 +2151,10 @@ no parameters.
...
+  Use `PascalCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
```
WizardOfOgz: Since both terms are used and understood separately by different people what do you think about using both names in the style guide?

``Use `UpperCamelCase` (PascalCase) for classes and modules.``

WizardOfOgz: Since both terms are used and understood separately by different people what do you think about using both names in the style guide?

``Use `UpperCamelCase` (PascalCase) for classes and modules.``

bbatsov: Fine by me.

jdickey: I like that. The distinction as "UpperCamelCase" (as opposed to PascalCase and camelCase) reminds the reader that this isn't your professor's camelCase being discussed here. :)

README.md
jimmcnulty41: Given the other compromise, I think this is a good idea

bbatsov:
@jimmcnulty41 Can you rebase this? We have a merge conflict right now.

README.md
bbatsov: Good point! Let's keep the name for now.

README.md
```diff
@@ -2151,10 +2151,10 @@ no parameters.
   ```
 
 
-* <a name="camelcase-classes"></a>
-  Use `CamelCase` for classes and modules.  (Keep acronyms like HTTP, RFC, XML
+* <a name="pascalcase-classes"></a>
+  Use `UpperCamelCase` (PascalCase) for classes and modules.  (Keep acronyms like HTTP, RFC, XML
```
bbatsov: I'd put `PascalCase` in backticks as well.

bbatsov:
I'm closing this PR due to no recent activity. Feel free to re-open it if you ever come back to it.

PR #647, opened 2017-03-21
bang-bang !! operator can be useful outside conditionals

aspiers:
Whilst the guide is correct to say that the `!!` double-bang operator is not appropriate within conditionals, I contend that in other contexts it can be a useful way to minimise unwanted leakage of values outside their intended scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1075,7 +1075,7 @@ Translations of the guide are available in the following languages:
   ```
 
 * <a name="no-bang-bang"></a>
-  Avoid the use of `!!`.
+  Avoid the use of `!!` in conditionals.
 <sup>[[link](#no-bang-bang)]</sup>
 
   `!!` converts a value to boolean, but you don't need this explicit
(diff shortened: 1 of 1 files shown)


webervin:
what about just either ` buggy_cleanup_method(!obj1.nil?, obj2)` or using TDD to ensure stuff works?

aspiers:
`!obj1.nil?` isn't too bad.  Another option might be to change the guide to recommend this as an alternative to `!!obj1` in that particular scenario.

Of course TDD is always a good idea but I don't think it should ever be considered a substitute for use of good style which encourages correctness, because we live in an imperfect world where test coverage is rarely 100%, and even when coverage is reported as 100%, it still won't guarantee to catch every single bug.

khiav223577:
I would prefer `obj1 != nil` over `!obj1.nil?` since both `!` and `.nil?` is negative.

aspiers:
`obj1 != nil` has the same number of negatives as `!obj1.nil?`.  I don't object to it, but it doesn't seem significantly different to me.  Arguably it's less idiomatic Ruby, since if direct comparisons to `nil` were considered idiomatic then there would be no reason for `.nil?` to exist.

pirj:
Closing as stale.

PR #650, opened 2017-04-13
Add section for named format tokens

backus:
This change matches the PR I've opened on RuboCop: https://github.com/bbatsov/rubocop/pull/4255


backus:
That was quick. Thanks @bbatsov

PR #652, opened 2017-04-29
Lowercase code block language identifier

a3ammar:
Hi,
This will allow Rouge and Github Pages to syntax highlight the code blocks used in this guide.

I want to turn this guide into a [Dash](https://kapeli.com/dash) document and trying to reuse the Github styles in the document. The simplest way to do that is to turn this guide into a Github Page and feed that into Dash. But due to capitalized language identifier used in the guide, Github Page doesn't highlight the code blocks. [Here's a live example.](https://a3ammar.github.io/ruby-style-guide/).


bbatsov:
👍 

Guess we'll have to do the same for the Ruby Style Guide for consistency.

PR #656, opened 2017-05-22
Use descriptive heredoc delimiters

Drenmi:
Using good delimiter names adds valuable information about the content of heredoc strings, makes it easier to match opening and closing tags, and as an added bonus many editors will highlight code enclosed with the correct delimiters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3803,28 +3803,57 @@ resource cleanup when possible.
 
   ```ruby
   # bad - using Powerpack String#strip_margin
-  code = <<-END.strip_margin('|')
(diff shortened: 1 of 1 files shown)


backus:
You should add a note about the syntax highlighting. That was one of the most compelling reasons for me

bbatsov:
> You should add a note about the syntax highlighting. That was one of the most compelling reasons for me

As that's editor-dependent I'd say readability is the most compelling argument. Still, it's worth mentioning this.

Drenmi:
@backus, @bbatsov: Added. 🙂

PR #657, opened 2017-05-24
Document consequences of module nesting

### LINKED ISSUE #637: Add note about nested modules declaration
I don't find any rules for nested modules declaration, such as
```
module Foo
  module Bar
    module Baz
      # some code
    end
  end
end
```
or
```
module Foo::Bar::Baz
  # some code
end

```

gareth:
Fixes #637

Note that #500 and #639 have already tried to solve this but haven't been merged. 

This example tries to use more relatable, real world examples, and adds an external link to a blog post with more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2614,6 +2614,42 @@ no parameters.
   end
   ```
 
+* <a name="namespace-definition"></a>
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #658, opened 2017-05-26
enforced handling of else case

thepeoplesbourgeois:
Chained ternary example to demonstrate more terse conditional syntax, with the added benefit of always handling the final `else` case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1007,15 +1007,44 @@ Translations of the guide are available in the following languages:
   result = some_condition ? something : something_else
   ```
 
+* <a name="else-case-enforcement"></a>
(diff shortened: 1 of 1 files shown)


bbatsov:
This seems anything but readable to me. :-)

thepeoplesbourgeois:
> This seems anything but readable to me. :-)

– Every newcomer to programming

bbatsov:
@thepeoplesbourgeois Everything is subjective in programming, no argument here. I see, however, no one supported this idea, so I guess we should close your PR. I kept it intentionally around for a while, as I was curious to see what others would say about it.

PR #667, opened 2017-07-16
Fix `method_missing` method definition example

diogotrentini:
Hi, everyone!

Just fixing `method_missing` definition example - it should not have a trailing `?`.


bbatsov:
👍

PR #669, opened 2017-07-31
Cleanup magic comments section

roberts1000:
The existing section is a little confusing so I attempted to clean it up.  The first rule talks about adding magic comments above all other code, but it also talks about putting them below shebangs.  My brain glossed over 

> Place magic comments above all code and documentation.

and went right to

> Magic comments should only go below shebangs if they are needed in your source file.

The very first code example, doesn't have anything to do with shebangs so I wasn't quite sure what I was looking at.  After a minute or two, I finally realized that the first example covers putting magic comments above everything else, and the second code example covered shebangs.

To make this a little more clear, I split the two thoughts into separate rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2469,32 +2469,39 @@ no parameters.
 ### Magic Comments
 
 * <a name="magic-comments-first"></a>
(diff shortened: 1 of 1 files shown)


bbatsov:
👍 Nice improvement!

PR #673, opened 2017-08-16
Update string literals quoting style with more descriptive examples

TeTiRoss:
This PR updating string literals quoting style examples, cause they are the same for single and double quoted cases, which I find confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3122,7 +3122,7 @@ no parameters.
     # ...other methods...
   end
   ```
-  
+
 ## Exceptions
 
 * <a name="prefer-raise-over-fail"></a>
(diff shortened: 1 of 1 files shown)


ufo2mstar:
Aww man! I was just about to make a similar change on the string literal example given :smile:  
haha, but guess you already got this.. thanks @TeTiRoss

bbatsov:
I think you missed the point of the examples - you're changes don't make any sense to me, unless you'd like to add some examples for exceptional situations.

TeTiRoss:
@bbatsov examples updated. I just want to make original examples a little different for two cases. Sorry, if I missed a point again :)

bbatsov:
Much better! Thanks!

PR #680, opened 2017-10-30
Update example for multi-line array alignment

bgeuken:
The ruby style guide now recommends using percent literals for arrays
of strings. This updates the examples accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -634,19 +634,19 @@ Translations of the guide are available in the following languages:
 
   ```ruby
   # bad - single indent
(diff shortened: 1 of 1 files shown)


ojab:
I think it's better to just mix `String` elements w/ something else (i. e. `Symbol`/`Numeric`) instead of using percent literals here.

bbatsov:
Yeah, I agree, but in its current form the consistency idea makes sense as well. Not a big deal.

PR #682, opened 2017-11-03
Update explanation for handling magic comments

bgeuken:
The previous explanation was a bit wordy and thus harder to
understand, eg. if there is no shebang, it should be clear that it
doesn't have to be considered:-)

Also it sounded like 'if they' would reference to magic comments, while
it actually meant a possible shebang.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2469,7 +2469,7 @@ no parameters.
 ### Magic Comments
 
 * <a name="magic-comments-first"></a>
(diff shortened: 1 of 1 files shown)


mikegee:
Consider:

> , but below the shebang in executable scripts.

to avoid implying the shebang is required in all Ruby files.

bbatsov:
Good point, @mikegee!

pirj:
This is pretty much covered in https://github.com/rubocop-hq/ruby-style-guide#below-shebang

PR #684, opened 2017-11-13
Allow a case of calling method with no argument parentheses

kachick:
Appending parentheses makes clarify method calling.
Especially `self.conflicted` does not fit for private methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -839,6 +839,30 @@ Translations of the guide are available in the following languages:
     puts temperance.age
     system 'ls'
     ```
+
+  Can use parentheses for
+
+  * Methods that have same name as local variables:
+
+    ```ruby
+    class FooBar
+      def foo
+        bar = :via_variable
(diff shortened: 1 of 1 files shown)


mikegee:
This is ok, but I think renaming either the local variable or the method is a better solution to this conflict. How you feel about making that recommendation explicit rather than accepting the parentheses?

bbatsov:
> Appending parentheses makes clarify method calling.

And is actually mandatory in such situations - that's why we didn't mention it in the guide to begin with. I like @mikegee's idea - you can mention this as a situation where parens are normally needed, but obvious better naming would solve that problem.

bbatsov:
Anyone willing to take this over?

pirj:
I'd rather not introduce a guideline that as an unwanted side effect would hint that using conflicting names is ok.
Do you guys mind if we just closed this?

PR #685, opened 2017-11-24
Colon method definition

rrosenblum:
This was requested in https://github.com/bbatsov/rubocop/pull/5100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -720,9 +720,9 @@ Translations of the guide are available in the following languages:
 ## Syntax
 
 * <a name="double-colons"></a>
-    Use `::` only to reference constants(this includes classes and
+    Use `::` only to reference constants (this includes classes and
     modules) and constructors (like `Array()` or `Nokogiri::HTML()`).
-    Do not use `::` for regular method invocation.
+     * Do not use `::` for regular method invocation.
 <sup>[[link](#double-colons)]</sup>
 
   ```ruby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -737,6 +737,23 @@ Translations of the guide are available in the following languages:
   SomeModule::SomeClass()
   ```
 
+  * Do not use `::` to define class methods.
```
bbatsov: You forgot the anchor here.

README.md
```diff
@@ -720,9 +720,9 @@ Translations of the guide are available in the following languages:
 ## Syntax
 
 * <a name="double-colons"></a>
-    Use `::` only to reference constants(this includes classes and
+    Use `::` only to reference constants (this includes classes and
     modules) and constructors (like `Array()` or `Nokogiri::HTML()`).
-    Do not use `::` for regular method invocation.
+     * Do not use `::` for regular method invocation.
```
bbatsov: Why did you convert this into a bullet point?

rrosenblum: Sorry, I should have explained my thought process. The first sentence sounds all encompassing about usages of `::`.
> Use `::` only to reference constants (this includes classes and
 modules) and constructors (like `Array()` or `Nokogiri::HTML()`).

It felt like ```Do not use `::` for regular method invocation``` and ```Do not use `::` to define class methods``` were supporting examples of where not to use double colons. I can see where this may not make sense, and I am open to suggestions on how to change this.

Issue #691, opened 2017-12-13
`&&=` is not reliable (maybe only for some versions)


EdmundLeex:
In 2.3.3

```ruby
[1] pry(main)> a &&= nil.a
=> nil
[2] pry(main)> a &&= 1.to_s
=> nil
[3] pry(main)> a = 1 && 1.to_s
=> "1"
[4] pry(main)> a &&= 1.to_s
=> "1"
[5] pry(main)> a &&= nil.a
NoMethodError: undefined method `a' for nil:NilClass
from (pry):5:in `__pry__'
```

Because of this, I wouldn't recommend using it.

EdmundLeex:
I have started an SO thread to discuss that. And I came to the conclusion of "don't use `&&=`".

pochmann:
Does exactly what I expected it to. I don't see any unreliability.

pirj:
Sorry, I don't really understand the lack of reliability in the application of `&&=`. It seems pretty consistent with what it does.
```ruby
x &&= y
```
is basically a shorthand for:
```ruby
x = y if x
```

PR #692, opened 2017-12-14
Don't use &&=

EdmundLeex:
Because it causes confusions and unexpected behaviors (by most people's standard).

https://stackoverflow.com/questions/47803567/in-ruby-operator-is-flaky#47803809
https://stackoverflow.com/questions/2982947/ruby-edge-case
http://www.rubyinside.com/what-rubys-double-pipe-or-equals-really-does-5488.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -1570,7 +1570,7 @@ condition](#safe-assignment-in-condition).
   ```
 
 * <a name="self-assignment"></a>
-  Use shorthand self assignment operators whenever applicable.
+  Use shorthand self assignment operators whenever applicable, except `&&=`.
 <sup>[[link](#self-assignment)]</sup>
 
   ```ruby
@@ -1588,7 +1588,7 @@ condition](#safe-assignment-in-condition).
   x **= y
   x /= y
   x ||= y
-  x &&= y
+  x = x && y
   ```
 
 * <a name="double-pipe-for-uninit"></a>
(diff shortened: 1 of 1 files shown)


pochmann:
Where's the evidence for "by most people's standard"?

EdmundLeex:
@pochmann Ok. I should say, in the style guide, it's suggesting using `a &&= b.c` to replace `a = b && b.c`. But they are inherently different.

EdmundLeex:
`a = b && b.c` blows up if `b` is not defined. Yet `a &&= b.c` assigns `nil` to `a`

pochmann:
> in the style guide, it's suggesting using `a &&= b.c` to replace `a = b && b.c`

What? Very very very much doubt it. Where exactly?

> Yet `a &&= b.c` assigns `nil` to `a`

Only if `a` wasn't defined before, and not because of the statement (but because Ruby creates the variable `a` and initializes it with `nil`).

pochmann:
But now it all makes sense. Why you think that [your examples](https://github.com/bbatsov/ruby-style-guide/issues/691) show `&&=` being unreliable and why you think that `a &&= nil.a` should always be `nil` and that `a &&= 1.to_s` should always be `'1'`. You **would** be right about all that if `a &&= b.c` indeed meant `a = b && b.c`. But it doesn't, and nobody but you thinks that it does.

Rule of thumb: Get your facts right before you try to tell everyone what they should and shouldn't do.

EdmundLeex:
@pochmann cool

PR #696, opened 2018-01-30
Add missing newline for consistency

Supernats:
While parsing this document, I discovered this is the one spot that does
not have a blank line before the next rule.


bbatsov:
👍

PR #700, opened 2018-02-08
Show bad examples before good ones

epistrephein:
Magic Comments section uses good examples before bad ones, unlike the rest of the guide.
This PR inverts the order of the examples, so that bad ones are shown first and guide style remains consistent.


bbatsov:
Thanks!

PR #706, opened 2018-03-01
Josh's suggested edits

jkolber:
Additional information about array sizing as it relates to ActiveRecord Relations.

Style suggestions for AASM transitions.

Created a list of methods that should trigger warnings when used.


jkolber:
My B, meant to open this against my fork

PR #720, opened 2018-06-15
Add example for coding in a functional way

ned:
We're cleaning out our internal style guide and thought that this example might be helpful for others.


bbatsov:
Thanks!

PR #721, opened 2018-06-19
[Fix #420] Recommend YARD instead of RDoc for API documentation

### LINKED ISSUE #420: TomDoc or Yard instead of Rdoc
The style-guide says:

```
Use RDoc and its conventions for API documentation.
```

However, the lack of consistent structure and readability in RDoc have lead to the specification of TomDoc. The style-guide should perhaps encourage TomDoc (optional)?

Here is some relevant reading:

http://tom.preston-werner.com/2010/05/11/tomdoc-reasonable-ruby-documentation.html
http://gnuu.org/2010/05/12/whats-missing-from-tomdoc/
http://tomdoc.org/

conradbeach:
The general consensus in issue #420 was that it would be preferable to recommend YARD instead of RDoc for API documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,8 +696,7 @@ Translations of the guide are available in the following languages:
   ```
 
 * <a name="api-documentation"></a>
-    Use [YARD][yard] and its conventions for API documentation.  Don't put an
-    empty line between the comment block and the `def`.
+    Use [YARD][yard] and its conventions for API documentation.
 <sup>[[link](#api-documentation)]</sup>
 
 * <a name="80-character-limits"></a>
--- a/README.md
+++ b/README.md
@@ -695,10 +695,10 @@ Translations of the guide are available in the following languages:
   num = 1234
   ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -695,10 +695,10 @@ Translations of the guide are available in the following languages:
   num = 1234
   ```
 
-* <a name="rdoc-conventions"></a>
-    Use [Rdoc][rdoc] and its conventions for API documentation.  Don't put an
+* <a name="api-documentation"></a>
+    Use [YARD][yard] and its conventions for API documentation.  Don't put an
```
bbatsov: Probably we can remove the second sentence here, as it seems unrelated to me.

bbatsov:
I'm fine with the change, although I wonder whether docs should be in the scope of the guide. Alternatively we can just remove this entry completely (I think this was suggested a few times as well).

README.md
conradbeach: Ok. I'll remove the second sentence.

Issue #359, opened 2014-09-03
Hanging Indentation on methods


baweaver:
I'd like to propose an addendum. Often times when one has a longer method, you have to do something with the parameters to get the method under 120 lines. One method of doing this is hanging indentation, which is very distracting when trying to grok a larger code base. It breaks the flow of reading.

``` ruby
# Bad - Hanging Indentation
some_method.with.a_longer.chain(:parameter_1,
                                :parameter_2,
                                :parameter_3,
                                :parameter_4)

# Good - Drop to a new line
some_method.with.a_longer.chain(
  :parameter_1,
  :parameter_2,
  :parameter_3,
  :parameter_4
)
```

Now granted the example is mildly confounded and you shouldn't chain methods that long. That being said, it's not unusual to see hanging indentations on 1-2 method chains past the 80th column which makes it near impossible to quickly skim over things.

Can we consider adding this to the guide?

weakish:
> One method of doing this is hanging indentation, which is very distracting
> when trying to grok a larger code base. It breaks the flow of reading.

@baweaver Another drawback of this is if you refactor the method name,
then all the following lines of parameters will be changed. This noises diff
if you are not using a custom diff.

> I don't recall having seen a similar convention in any other language I've
> ever worked in.

@jdickey There is one in [scheme][]

[scheme]: http://community.schemewiki.org/?scheme-style

```scheme
 (+ 1 
    (foo 3.5 a) 
    bar 
    baz) 
```

> Our shop standard allows for doubly-indented continuation lines

@jdickey  What about using one level of indentation?

```ruby
some_method.with.a_longer_chain :parameter_1,
  :parameter_2,
  :parameter_3,
  :parameter_4

# An empty line above more code...
```

> But I think we already have [similar rules in the guide][guide].

[guide]: https://github.com/bbatsov/ruby-style-guide#no-double-indent

@bbatsov  That rules does not need to consider indentation.

jdickey:
@weakish The reason we prefer two levels of indentation to one in that usage is to avoid confusion with a block, such as for a conditional. With one level of indentation, you'd be skimming down the listing, see things go one level in, and go back and look for a conditional or such that you missed the first time. "Don't overload the meaning of a single indent", is how one dev put it.

I can live with the single indent; it's just not so immediately clear what it *means*.

strand:
[...]

PR #725, opened 2018-08-15
#359 - No Hanging Indentation

baweaver:
Based off of a very old issue that had a lot of interesting discussion:

https://github.com/rubocop-hq/ruby-style-guide/issues/359

The idea here is that hanging indentation is incredibly hard to read while going down a page, and also incurs extra editing whenever more methods are added to a chain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -624,14 +624,6 @@ Translations of the guide are available in the following languages:
         body: source.text)
   end
 
-  # good
-  def send_mail(source)
-    Mailer.deliver(to: 'bob@example.com',
-                   from: 'us@example.com',
-                   subject: 'Important message',
-                   body: source.text)
-  end
-
   # good (normal indent)
   def send_mail(source)
     Mailer.deliver(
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -643,6 +635,40 @@ Translations of the guide are available in the following languages:
...
   ```
 
+* <a name="no-hanging-indent"></a>
+    Use a new line for the argument list of a method at the end of a chain
+    if they are indented more than one level from the first method.
+<sup>[[link](#no-hanging-indent)]</sup>
+
+  ```ruby
+  # bad (hanging indent)
+  some_method.with.a_longer.chain(:parameter_1,
```
bbatsov: I think this needs to be made clear with respect to:

1) no method chain  - e.g. some_method(a,
2) short method chain - frankly it seems to the example you deleted looks fine, as it's not really a long method chain.

README.md
```diff
@@ -643,6 +635,40 @@ Translations of the guide are available in the following languages:
   end
   ```
 
+* <a name="no-hanging-indent"></a>
```
bbatsov: Maybe `no-hanging-param-indent` or something like this?

PR #728, opened 2018-09-15
Update operand name rule for binary operators.

zverok:
Make it mentioning more real-life considerations.

As per discussion in Rubocop: https://github.com/rubocop-hq/rubocop/pull/6295#issuecomment-421537232

### DIFF
--- a/README.md
+++ b/README.md
@@ -2194,7 +2194,7 @@ no parameters.
   :some_sym_1
 
   some_var_1 = 1
-  
+
   var_10  = 10
 
   def some_method_1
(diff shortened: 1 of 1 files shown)


bbatsov:
Looks good to me. Just reorganize the examples so there's just one `good` and `bad` group for consistency with the rest of the guide.

bbatsov:
Actually I noticed the guide itself is pretty inconsistent in this regard right now. A point for improvement. :-)

zverok:
Looks like we need style guide for style guide :trollface: 
And probably some styleguideocop to enforce it!

PR #729, opened 2018-09-18
update markdown indention

3limin4t0r:
Changed indention from 2 to 4 spaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,36 +61,36 @@ style guide.
 
 Translations of the guide are available in the following languages:
 
-* [Chinese Simplified](https://github.com/JuanitoFatas/ruby-style-guide/blob/master/README-zhCN.md)
-* [Chinese Traditional](https://github.com/JuanitoFatas/ruby-style-guide/blob/master/README-zhTW.md)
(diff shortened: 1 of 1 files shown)


bbatsov:
I guess you've checked that this solved the problem you reported originally, right?

bbatsov:
This seems to indicate you're right http://brettterpstra.com/2015/08/24/write-better-markdown/#indentation

We'll likely have to do something similar with the other guides as well. //cc @Darhazer @pirj

3limin4t0r:
> I guess you've checked that this solved the problem you reported originally, right?

@bbatsov Yes, I check if this resolved the issue. I attached the correctly generated PDF to the issue in my last comment (https://github.com/rubocop-hq/ruby-style-guide/issues/727#issuecomment-422367549). This also includes the command to generate the PDF so you can test it for yourself if you'd like.

bbatsov:
Thanks for tackling this! 

If you have some time and desire - you might also fix our Rails and RSpec guides as well. :-)

3limin4t0r:
@bbatsov I've created PRs for them, see the references above.

PR #730, opened 2018-10-01
Unban `and` and `or` control flow operators

zverok:
And so, after several years of private discussions, here I am!

My reasoning goes as this:

1. I remember/understand where the current state of the rule came from. Initially, `and`/`or` were understood by most of the community (and previous versions of this guide) as "broken logical operators" with surprisingly low priority, and, to add insult to injury, equal priority towards each other (unlike usual higher priority of logical "and"). But, as time went and approaches to Ruby style and coding practices became crystallized, it became obvious that `and`/`or` were designed for **control flow**, and their design is extremely suitable for this task, as shown by examples below:
  ```ruby
  # parsing and evaluation priority allows to use and/or as clear precondition for methods:
  value = fetch_necessary_value or raise Error, "Pretty long error message"
  value = fetch_optional_value or return

  # their different spelling and priority are good for mixing them with logical evaluation
  possible_error_check || other_possible_error && condition_for_that_error and raise
  # There is **no** clear and DRY replacements for examples above.
  ```
2. This style guide went a long road since its inception. In current state, it is de-facto "**The** Style Guide", one that novices are immediately exposed to, one that The Linter (aka Rubocop) implements with default settings. With this in mind, stakes for "doing right" in style guide are much higher than their were yeas ago, it is a lot of responsibility for entire community/infrastructure. "Banning" of useful, clearly designed language feature with readable usages and obvious meaning seems just plain wrong—**despite** possible discussions of who prefers what.
3. It is important to observe that the style guide does NOT, in current state, totally against flow control with "and"/"or", but advices (inappropriately) `&&` and `||` for it, which do not allow most of the good usages of flow control! It is slightly like "screwdrivers are banned; use your pocket knife to screw things".
4. Aside observation: no other language feature is "banned". Some are recommended to "avoid", most of the time with explanantion _why_ and/or recommendations _what_ should be used instead; the esoteric flip-flops are almost the only thing that just commented with "avoid, period", but even `for` recommendation is softened with "unless you know exactly why". So, this "personal hostility" towards this particular feature looks obviously outdated.
5. The last objection to `and`/`or` that could be thought of is their equal priority. I believe that it doesn't matter at all, as for the flow control more than one flow-control operator in a row is weird anyways (and my proposed change advices on that). _As a one more side note, I assume that idea of equal priority came exactly from "flow control" domain: to avoid "reordering" of flow statements, in examples like: `output or log and exit` ("regular" priorities would read it as `output or (log and exit)`, so no `exit` if `output` was successful, "flat" priorities would do `(output or log) and exit`). I agree it **could** be thought as convoluted, and, again, advice against combining flow control operators._

I believe this justifies the change proposed.

**NB**: I am preserving `#no-and-or-or` anchor (in addition to the new `#and-or-flow` anchor), in order to old links to this rule still work (but leading to the new rule).

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1420,57 +1420,63 @@ def banned?
 end
 ----
 
-=== `and`/`or` [[no-and-or-or]]
+=== `and`/`or` [[no-and-or-or]] [[and-or-flow]]
 
-The `and` and `or` keywords are banned.
-The minimal added readability is just not worth the high probability of introducing subtle bugs.
-For boolean expressions, always use `&&` and `||` instead.
-For flow control, use `if` and `unless`; `&&` and `||` are also acceptable but less clear.
+`and` and `or` are control flow operators. They have very low precedence, and should only be used
+as a final fallback statements.
 
 [source,ruby]
 ----
 # bad
-# boolean expression
-ok = got_needed_arguments and arguments_are_valid
-
-# control flow
-document.save or raise("Failed to save document!")
+# and/or in conditions
+if got_needed_arguments and arguments_valid
+  # ...body omitted
+end
+# in logical expression calculation
+ok = got_needed_arguments and arguments_valid
 
 # good
-# boolean expression
-ok = got_needed_arguments && arguments_are_valid
+# &&/|| in conditions
+if got_needed_arguments && arguments_valid
+  # ...body omitted
+end
+# in logical expression calculation
+ok = got_needed_arguments && arguments_valid
 
-# control flow
-raise("Failed to save document!") unless document.save
+# bad
+# &&/|| for control flow (can lead to very surprising results)
+x = extract_arguments || raise(ArgumentError, "Not enough arguments!")
 
-# ok
-# control flow
-document.save || raise("Failed to save document!")
+# good: and/or for control flow
+x = extract_arguments or raise ArgumentError, "Not enough arguments!"
+user.suspended? and return :denied
 ----
 
-.Why Ban `and` and `or`?
-****
-The main reason is very simple - they add a lot of cognitive overhead, as they don't behave like similarly named operators in other languages.
-
-First of all, `and` and `or` operators have lower precedence than the `=` operator, whereas the `&&` and `||` operators have higher precedence than the `=` operator, based on order of operations.
+But avoid several control flow operators in one expression, that becomes
+confusing:
 
 [source,ruby]
 ----
-foo = true and false # results in foo being equal to true. Equivalent to ( foo = true ) and false
-bar = false or true  # results in bar being equal to false. Equivalent to ( bar = false ) or true
-----
+# bad
+# Did author mean conditional return because `#log` could result in `nil`?
+# ...or was it just to have a smart one-liner?
+x = extract_arguments and log("extracted") and return
 
-Also `&&` has higher precedence than `||`, where as `and` and `or` have the same one. Funny enough, even though `and` and `or`
-were inspired by Perl, they don't have different precedence in Perl.
+# good
+# If the intention was conditional return
+x = extract_arguments
+if x
+  return if log("extracted")
+end
+# If the intention was just "log, then return"
+x = extract_arguments
+if x
+  log("extracted")
+  return
+end
+---
 
-[source,ruby]
-----
-foo = true or true and false # => false (it's effectively (true or true) and false)
-foz = true || true && false # => true (it's effectively true || (true && false)
-bar = false or true and false # => false (it's effectively (false or true) and false)
-baz = false || true && false # => false (it's effectively false || (true && false))
-----
-****
+NOTE: whether organizing control flow with `and` and `or` is a good idea was a controversial topic in community for a long time. But if you do, use these operators and not `&&`/`||`.
 
 === Multi-line Ternary Operator [[no-multiline-ternary]]
 
--- a/README.adoc
+++ b/README.adoc
@@ -1422,13 +1422,19 @@ end
 
 === `and`/`or` [[no-and-or-or]] [[and-or-flow]]
 
-`and` and `or` are control flow operators. They have very low precedence, and should only be used
-as a final fallback statements.
+`and` and `or` are control flow operators. They have very low precedence, and can be used as a short
+form of specifying flow sequences like "evaluate expression 1, and only if it is not successful
+(returned `nil`), evaluate expression 2". This is especially useful for raising errors or early
+return without breaking the reading flow.
 
 [source,ruby]
 ----
+# good: and/or for control flow
+x = extract_arguments or raise ArgumentError, "Not enough arguments!"
+user.suspended? and return :denied
+
 # bad
-# and/or in conditions
+# and/or in conditions (their precedence is low, might produce unexpected result)
 if got_needed_arguments and arguments_valid
   # ...body omitted
 end
@@ -1446,10 +1452,6 @@ ok = got_needed_arguments && arguments_valid
 # bad
 # &&/|| for control flow (can lead to very surprising results)
 x = extract_arguments || raise(ArgumentError, "Not enough arguments!")
-
-# good: and/or for control flow
-x = extract_arguments or raise ArgumentError, "Not enough arguments!"
-user.suspended? and return :denied
 ----
 
 But avoid several control flow operators in one expression, that becomes
@@ -1476,7 +1478,7 @@ if x
 end
 ---
 
-NOTE: whether organizing control flow with `and` and `or` is a good idea was a controversial topic in community for a long time. But if you do, use these operators and not `&&`/`||`.
+NOTE: Whether organizing control flow with `and` and `or` is a good idea was a controversial topic in the community for a long time. But if you do, prefer these operators over `&&`/`||`.
 
 === Multi-line Ternary Operator [[no-multiline-ternary]]
 


mikegee:
You make some strong arguments, but I'm still not convinced. Your best examples of their usage (under bullet `1.` above), are not strictly superior to the 2-line versions of those expressions: 

```ruby
# value = fetch_necessary_value or raise Error, "Pretty long error message"

value = fetch_necessary_value
raise Error, "Pretty long error message" unless value
# or :)
value || raise Error, "Pretty long error message"
```
```ruby
# value = fetch_optional_value or return

value = fetch_optional_value
return unless value
# or :)
value || return
```
```ruby
# possible_error_check || other_possible_error && condition_for_that_error and raise

problem = possible_error_check || other_possible_error && condition_for_that_error
raise if problem
# or :)
problem && raise
```
While I'm not comfortable banning part of Ruby on principle either, I haven't seen code that was improved with `and`/`or` over many years of full-time Ruby development.

bbatsov:
I'll not comment on this right away. I'd rather like to hear what @rubocop-hq/rubocop-core members think, @pirj, and everyone else who'd like to participate of course.

pirj:
I'm bad at remembering things and prefer simple code that I wouldn't have to spend much time understanding, and preferably wouldn't have to look up operator order precedence in the docs.

The usage of `or` and `and` is quite uncommon, and I would prefer that they won't appear as the new thing to learn, and so I wouldn't have to remember it.

It's not correct that the guide advises using `||`/`&&` instead of `or`/`and` in control statements:
> For flow control, use `if` and `unless`

It's totally up to the project maintainers to turn off this or any other rule if they decide the developers working with the code on a regular basis are well aware of how exactly this works and should be used.

Would you agree with softening the wording from
> The `and` and `or` keywords are banned.
to
> Avoid the use of `and` and `or` keywords.
?

The use of one-liners is justified when the code is clean and readable. I'm afraid the example with a very long error message and a conditional doesn't fall into this category.
The below is my personal preference, I'm not putting it here as a reference example of how any of the snippets above should be rewritten, just an option to consider:
```
value = fetch_necessary_value
raise OurNamespace::SomeParticularError, <<~MESSAGE unless value
  Expected a value to be fetched when #{context},
  but there were none. Try providing a value.
MESSAGE
```

jonas054:
What if we limit the use of `and` and `or` to expressions where the right hand side is a "flow command" such as `return`, `break`, or `exit`? Would there still be a risk for misunderstanding of function and subtle bugs? Give me an example if you can come up with one.

This is something we could easily check in RuboCop, BTW.

zverok:
OK, let me be bitter. 

So far, all counter-arguments can be boiled down to exactly two statements:
1. **Usage of `and`/`or` may produce "subtle bugs".** No examples were provided here to support this claim, but I believe that it is about using of those operators in boolean context (`if`s and calculation of boolean variables). Which can be properly rephrased into: "This (sane) language feature, when used in a wrong way, produces bugs". Well... Ain't all of them? IDK, shouldn't we ban bit `&`, because if somebody  uses it instead of `&&`... BAD things will happen. 
2. **"I don't like it"** (Don't see any value, don't think it is readable, or even, like the guide currently say, in a more cunny way "The _minimal_ added readability doesn't worth the _high_ probability..." -- argree that readability is "added", but make it look neglectible).

And (2) is **exactly** the problem I am trying to address. I believe that `and`/`or` operators fell into "negative feedback loop": borrowed from Perl, first, they were thought of by most Rubyists as "broken boolean ops"; then they were forgotten/advised against; then nobody has used them; then, if you try to tell someone "they are good _for what they designed_", people are like "ah, never seen them and I remember there were some bad things about". And nobody sees them because of it, and everybody just thinks "I never seen it" :)

In other words, people tend to constantly confuse "I haven't tried to use it" with "it is unreadable / not clean code".

The thing is (and I'll be bold here, the discussion seems to be in a dead-end anyways), `var = something and/or raise/return` idiom can be **proven to be better** easily:
* it is DRY (don't repeat the variable you've just calculated to check if it was good)—Rubyists are used to not repeating names, that's where `Symbol#to_proc` came from
* it is in a spirit of Ruby's "chaining statements related to the same value" (besides dot-chain you can also remember how popular even if questionable AS's `.presence` is)
* it is naturally readable and rememberable: I am saying it from experience with dozens of students I've exposed to this idiom, the typical reaction was intuitive understanding and joy
* it is handy to write/read (my example with raise and long error message)
* as any statement chaining, it can lead to better code (when trying to express meaning so that it would allow using this idiom, you will in many cases extract pre-checks/calculations of method arguments, indirectly making method's body readability higher).

But however long I'll make the list above, the answers would be "empirical": I don't see how it is readable, I don't like how it looks, I never saw it used in production, Python doesn't have it and it would be confusing, CoffeeScript uses `and` for logical operations and it would be confusing and so on. All saying mostly the same: _**I** don't use it, so there **should** be something wrong with it._

Unfortunately, at the end of the day, I am maybe too old and disillusioned to fight those windmills. And Ruby's dying anyways :)

bquorning:
> The usage of `or` and `and` is quite uncommon, and I would prefer that they won't appear as the new thing to learn, and so I wouldn't have to remember it.

I wonder if not RuboCop is at least partially to blame for that.

`Style/AndOr` is one of the few cops that I *always* disable on a project. I write Ruby, I have read [@avdi’s article](http://www.virtuouscode.com/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/), and I know how `and` and `or` works. I don’t like that the community style guide bans valid language features.

On the other hand, you don’t have to look very far to find popular projects using control flow keywords as boolean operators. The first three gems I checked gave three results:

  - https://github.com/rack/rack/blob/5559676e7b5a3107d39552285ce8b714b672bde6/lib/rack/request.rb#L465
  - https://github.com/carrierwaveuploader/carrierwave/blob/f32db18a3cd45a55d4e008990a814438a959ed6e/lib/carrierwave/uploader/url.rb#L18
  - https://github.com/sinatra/sinatra/blob/b7c10649cad56f3263df089027621fb1753620cb/lib/sinatra/base.rb#L216

So while I may claim to always only use `and` and `or` for control flow, evidently they are often used as boolean operators.

matkoniecz:
> as time went and approaches to Ruby style and coding practices became crystallized, it became obvious that and/or were designed for control flow

I would dispute it. Maybe this use became obvious for experts, but it is still not obvious for nonexperts.

matkoniecz:
Also, I would not consider

> # their different spelling and priority are good for mixing them with logical evaluation
> possible_error_check || other_possible_error && condition_for_that_error and raise

as a readable code

3limin4t0r:
@matkoniecz That is because you're doing too much in one same statement. You can make anything unreadable if you want to. At least it's clear that if `# insert some logic` evaluates truthy then an exception will be rasied. However, that might indeed not be the best example. But I find the variable assignment pretty readable.

```Ruby
# 1.
variable = something_that_might_return_falsey or return

# 2. 
return unless (variable = something_that_might_return_falsey)

# 3. 
variable = something_that_might_return_falsey
return unless variable
```

In my opinion option 2 is just plain bad. Whereas option 1 and 3 are both equally readable. For this reason I would go for option 1 since it's one line less. If you follow the default RuboCop guidelines a method may only have 10 lines of content. Changing two lines into one is pretty huge if you don't sacrifice any readability.

matkoniecz:
For me:

2 is incredibly ugly and unclear.
1 is unclear
3 has one more line and is not using confusing constructs

I would save 2 style for code golf puzzles

README.md
```diff
@@ -1200,31 +1200,58 @@ Translations of the guide are available in the following languages:
...
+    x = extract_arguments
+    if x
+      return if log("extracted")
+    end
+    # If the intention was just "log, then return"
+    x = extract_arguments
+    if x
+      log("extracted")
+      return
+    end
```
ka8725, line 1254: What about mix of logical and flow operator like the following?

```ruby
this.is_not_right? and raise 'error' unless may_be_raise?
```

Is that good or bad? If bad, what's good alternative?

mvz: IMHO, it's bad because there are too many control flow keywords there, just like:

```
raise 'error' if this.is_not_right? unless may_be_raise?
```

A good alternative:
```
unless may_be_raise?
  this.is_not_right? and raise 'error'
end
```

(It seems the logic is inverted for the name `may_be_raise?`)

zverok:
@bbatsov Is it possible to make, 4 years later, an executive decision and acknowledge that banning of language features was an early mistake of the style guide?.. The current situation is really honestly weird. The [corresponding cop](https://docs.rubocop.org/rubocop/1.32/cops_style.html#styleandor) currently does the following:
* has `conditionals` as the default setting (directly contradicting the bold Style Guide's statement "The `and` and `or` keywords are banned. ")
* suggests that `&&` is cool for control flow (which it is NOT)
```ruby
# Cop's example:
# good
foo.save && return

# does it look good for you?..
render 'my_template' && return # parsed as render('my_template' && return)
# does it look good for you?
user.suspended? && return :denied # Syntax error
```

README.md
```diff
@@ -1200,31 +1200,58 @@ Translations of the guide are available in the following languages:
...
     ```
 
-  * <a name="no-and-or-or"></a>
-    The `and` and `or` keywords are banned. The minimal added readability is just
-    not worth the high probability of introducing subtle bugs. For boolean
-    expressions, always use `&&` and `||` instead. For flow control, use
-    `if` and `unless`; `&&` and `||` are also acceptable but less clear.
-    <sup>[[link](#no-and-or-or)]</sup>
+  * <a name="no-and-or-or"></a><a name="and-or-flow"></a>
+    Distinguish between `&&`/`||` logical operators and `and`/`or` control flow
```
bbatsov, line 1204: I'd phrase this more like a rune - e.g. Don't use `and` and `or` as logical operators...

bbatsov:
@zverok Yeah, I will - let's allow them, but let's be more specific on the wording and add extra rationale (e.g. maybe a big sidebar and some note this has long been a controversial topic in the Ruby community).

zverok:
@bbatsov Thanks. I rebased and reworded the PR on the new README.adoc, please take a look :pray:

README.adoc
```diff
@@ -1420,57 +1420,63 @@ def banned?
...
 
-=== `and`/`or` [[no-and-or-or]]
+=== `and`/`or` [[no-and-or-or]] [[and-or-flow]]
 
-The `and` and `or` keywords are banned.
-The minimal added readability is just not worth the high probability of introducing subtle bugs.
-For boolean expressions, always use `&&` and `||` instead.
-For flow control, use `if` and `unless`; `&&` and `||` are also acceptable but less clear.
+`and` and `or` are control flow operators. They have very low precedence, and should only be used
+as a final fallback statements.
```
bbatsov, line 1426: Why fallback? I guess you meant final expression in a method in general, right?

README.adoc
```diff
@@ -1420,57 +1420,63 @@ def banned?
...
 
-[source,ruby]
-----
-foo = true or true and false # => false (it's effectively (true or true) and false)
-foz = true || true && false # => true (it's effectively true || (true && false)
-bar = false or true and false # => false (it's effectively (false or true) and false)
-baz = false || true && false # => false (it's effectively false || (true && false))
-----
-****
+NOTE: whether organizing control flow with `and` and `or` is a good idea was a controversial topic in community for a long time. But if you do, use these operators and not `&&`/`||`.
```
bbatsov, line 1479: whether -> Whether
use these -> prefer these operators over

README.adoc
zverok: @bbatsov I tried to adjust the wording, making it more explanatory. I also moved the "good usage" example to be the first one, so it also works as a demo of the concept.

bbatsov (approved):

zverok:
:tada: 
I remember when we first met (on Kyiv RubyC in 2017? probably) my very first phrases was: "Hi, nice to meet you! BTW, Rubocop is cool, but why are you so severe to `and`/`or`? They are cool too!"
Mission accomplished. :grin: 

Thanks.

bbatsov:
Good things take a while... 😉

WaKeMaTTa:
This is not good. 

`and` and `or` should kept banned, because is confusing for beginners.

PR #734, opened 2019-01-09
Clarify when to use if-case returns

wrs:
There are two contradictory guidelines that shouldn't really be contradictory.

* Favor the ternary operator(`?:`) over `if/then/else/end` constructs.
* Leverage the fact that `if` and `case` are expressions which return a result.

This adds a little more color on when to use one or the other.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1090,8 +1090,9 @@ Translations of the guide are available in the following languages:
     ```
 
   * <a name="ternary-operator"></a>
-    Favor the ternary operator(`?:`) over `if/then/else/end` constructs.
-    It's more common and obviously more concise.
+    Favor the ternary operator(`?:`) over `if/then/else/end` constructs
+    for simple expressions. It's more common and obviously more
+    concise.
     <sup>[[link](#ternary-operator)]</sup>
 
     ```ruby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1134,8 +1135,10 @@ Translations of the guide are available in the following languages:
     ```
 
   * <a name="use-if-case-returns"></a>
-    Leverage the fact that `if` and `case` are expressions which return a
-    result.
+    When assigning a value that depends on a conditional, when
```
bbatsov: I agree with the premise, but not with the wording. This long sentence is very hard to process and should probably be broken down into 2-3 sentences.

wrs: It seems like this is really two rules:

(a) When doing an assignment that depends on a conditional, don't repeat the assignment statement in an if/then, use a conditional expression instead.

(b) When assigning a conditional expression, choose to use a ternary operator or an if/then expression based on the complexity of the expression.

I'm not sure the best way to accomplish that. I'd suggest restructuring the section to have two rules just like (a) and (b), but I'm assuming it's hard to change existing rules since Rubocop and others may use the anchor names. (No?)

PR #736, opened 2019-01-09
Mention ActiveSupport::Inflector

wrs:
The guideline to keep acronyms uppercase in CamelCase means people run into problems with Rails autoloading (failed autoload or a file called `x_m_l_something`). Link to how to avoid that.


bbatsov:
We've avoided on purposes referring to any third-party libraries in the style guide.

PR #738, opened 2019-01-16
Fixed a small styling problem at the "Syntax" chapter

SilasSpet:
The indentation of the code snippet before "Can omit parentheses for" was wrong which made that whole part look like a code snippet. This fixes the styling again.


koic:
Thanks!

PR #422, opened 2015-03-25
Multi condition case when

NewAlexandria:
Depends on _Chanel #9_

Following-up from the PR on `When/Then` block

There is no demonstration to confirm how to format when multiple conditions should result in the same action.  I propose this:

``` Ruby
# good

case token
when :star_op
  stack.pop * stack.pop
when :slash_op
  stack.pop / stack.pop
when :minus_op, :minus_minus_op
  also_calculate_that
  stack.pop - stack.pop
when MyModule::SomeDomain::BETA_USERS,
     MyModule::SomeDomain::INTERNAL_RELEASE
   stack.pop + stack.pop
when :int_literal,
     :some_complicate_explicit_name,
     :graduate_borrowers_with_arms,
     :str_interpolated
  token.value
end
```

Though better control of primary domain references should be exercised, this style offers a solution for some 'in the wild' situations.   It reads better than:

``` Ruby
# bad

case token
when :star_op
  stack.pop * stack.pop
when :slash_op
   stack.pop / stack.pop
when :minus_op, :minus_minus_op
  also_calculate_that
  stack.pop - stack.pop
when MyModule::SomeDomain::BETA_USERS, MyModule::SomeDomain::INTERNAL_RELEASE
  stack.pop + stack.pop
when :int_literal, :some_complicate_explicit_name, :graduate_borrowers_with_arms, :str_interpolated
  token.value
end
```

Where the 'bad' example also has the issue of cause the entire when line to diff when one of the conditions is changed or updated.  

I know that some people love long lines, but I suggest that a newline is an easier-to-cognized demarcation of conditions, rathe than a comma

PR #741, opened 2019-03-17
Better clarity on case-when indentation

NewAlexandria:
This replaces PR #422, and maintains a minimum of overall new lines.

----

I just realized that #422 was still open, and re-reading it I think that this edit is more in-line with the intent of the PR, and all the feedback.  I hope it will an easy merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,8 +327,12 @@ Translations of the guide are available in the following languages:
         puts 'Not again!'
       when song.duration > 120
         puts 'Too long!'
-      when Time.now.hour > 21
-        puts "It's too late"
+      when :minus_op, :minus_minus_op
+        stack.pop - stack.pop
+      when MyModule::SomeDomain::BETA_USERS, MyModule::SomeDomain::INTERNAL_RELEASE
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm fine with the spirit of the change, I just wonder if this should be a separate rule or not. It's kind of orthogonal to the indentation of `when` itself.

README.md
```diff
@@ -341,11 +345,26 @@ Translations of the guide are available in the following languages:
       puts 'Too long!'
     when Time.now.hour > 21
       puts "It's too late"
+    when :minus_op, :minus_minus_op
```
bbatsov: Given this example, I think we're talking more about `Good` and `Better` than about `Bad` and `Good`.

NewAlexandria: There's just one other precedent in the style guide.  I'm happy to make a simple `Better` block, and try to overall keep the examples 'deduped' to limit newlines.

NewAlexandria: Updated @bbatsov.   A smaller diff now, too.

PR #746, opened 2019-04-30
Add heredoc argument closing parenthesis note

maxh:
Let's add a note about this cop to the style guide? https://github.com/rubocop-hq/rubocop/pull/6932


bbatsov:
Thanks! You can also link to the rule from the cop now that it exists.

PR #748, opened 2019-06-03
Convert to AsciiDoc

pirj:
AsciiDoc allows us to have native admonitions, footnotes, sane nesting of list items, etc and to easily export the guide in various formats.  That would also make it trivial to publish a simple site from the guide down the road if we want to.

A longer term plan is to convert all guides to AsciiDoc and maybe even RuboCop's own documentation. RSpec and Rails style guides have been recently converted:

  - https://github.com/rubocop-hq/rspec-style-guide/pull/97
  - https://github.com/rubocop-hq/rails-style-guide/pull/250

Easy view of the resulting file: https://github.com/rubocop-hq/ruby-style-guide/blob/convert-to-asciidoc/README.adoc


bbatsov:
Fantastic work, @pirj! 🙇 

Now we're left only with the conversion of the manual to Antora + AsciiDoc, but that's going to be trickier, because we generate a portion of it programatically (the listing of all cops). I'll create some ticket on RuboCop's issue tracker for this at some point. 

I'll also think how we can publish the guides to some static site.

PR #750, opened 2019-06-04
Quote code in section titles

pirj:
During recent AsciiDoc conversion, section titles were introduced, and some of those titles are suboptimal in terms of readability, e.g.

    Use if case returns [[use-if-case-returns]]

This change improves some of them.

It's ok to change titles as long as original section anchor is kept.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -290,7 +290,7 @@ No space inside range literals.
 'a'...'z'
 ----
 
-=== Indent when to case [[indent-when-to-case]]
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

> During recent AsciiDoc conversion, section titles were introduced, and some of those titles are suboptimal in terms of readability, e.g.

Yeah, I've noticed this as well. Generally those titles should be more descriptive than the anchors, so there's definitely room for improvement with them.

PR #751, opened 2019-06-09
Use proper Title Case in section titles

pirj:
Section titles were mostly auto-generated during AsciiDoc conversion, and some of them were off, e.g. "Single line" vs "Single-line".
But the main culprit that they were not using proper Title Case.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -73,7 +73,7 @@ ____
 
 Use `UTF-8` as the source file encoding.
 
-=== Spaces indentation [[spaces-indentation]]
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks! That's a great step in the right direction, but we should also reword many of the titles in general to have more meaningful names. E.g. - "Spaces Indentation" should probably be named just indentation (and be split into two rules), etc. Ideally at some point we should have alignment between the names of the rules in the guide and the cops that enforce them in RuboCop.

pirj:
Thanks!
Looking forward to it. 👍

PR #752, opened 2019-06-11
Add `when` description/example to `no-then` rule

okuramasafumi:
There's a PR to add a cop named `MultilineWhenThen`,
which detects `then` in multiline `when`.
https://github.com/rubocop-hq/rubocop/pull/7114
It refers to `MultilineIfThen` cop, which is based on `no-then` rule here.
So this commit adds `when` description and example to `no-then` rule,
making newly added `MultilineWhenThen` cop based on the style guide.
FYI, the discussion about this feature is here: https://github.com/rubocop-hq/rubocop/issues/7064

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1081,7 +1081,7 @@ elem # => NameError: undefined local variable or method `elem'
 
 === No `then` [[no-then]]
 
-Do not use `then` for multi-line `if`/`unless`.
+Do not use `then` for multi-line `if`/`unless`/`when`.
 
 [source,ruby]
 ----
@@ -1090,8 +1090,20 @@ if some_condition then
   # body omitted
 end
 
+# bad
+case foo
+when bar then
+  # body omitted
+end
+
 # good
 if some_condition
+  # body omitted
+end
+
+# good
+case foo
+when bar
   # body omitted
 end
 ----


README.adoc
```diff
@@ -1092,7 +1092,7 @@ elem # => NameError: undefined local variable or method `elem'
 
 === No `then` [[no-then]]
 
-Do not use `then` for multi-line `if`/`unless`.
+Do not use `then` for multi-line `if`/`unless` or `when`.
```
koic: It seems that `/` is conventionally used as a separator.

```diff
- Do not use `then` for multi-line `if`/`unless` or `when`.
+ Do not use `then` for multi-line `if`/`unless`/`when`.
```

koic:
I'm not sure if names (`some_var `, `some_val`) shown in added example code are suitable. Anyway I think that it is consistent as a "No `then`" style.

README.adoc
okuramasafumi: True, I'll fix it.

okuramasafumi:
@koic I noticed that `some_var` and `some_val` can be replaced with better words, but I couldn't come up with them.

mikegee:
Consider something like:

```ruby
case x
when 42 then
  # body omitted
end
```

bbatsov:
`foo` and `bar` are always options as well. ;-)

Btw, you'll need to rebase this, as I've been shuffling things around today to improve a bit the structure of the guide.

okuramasafumi:
@bbatsov I found `foo` and `bar` is clearer :)

bbatsov:
Thanks!

PR #755, opened 2019-06-12
Add rule for super method with parens

### LINKED ISSUE #703: The super method, parentheses with arguments?
For the `super` method being a keyword in ruby, I am curious as to if it requires parentheses around arguments or not. I am having trouble grasping if it is declarative or not.

From my understanding these are correct:

```
def something_1
  super
end

def something_2
  super(:foo, :bar, :baz)
end

def something_3
  super :foo, :bar, :baz
end
```
 In terms of `something_2` vs `something_3` is one better over the other?

tejasbubane:
Closes #703.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2442,6 +2442,21 @@ puts temperance.age
 system 'ls'
 ----
 
+==== "super" method usage [[super-parens]]
+
+Always use parentheses when calling `super` with arguments:
+
+[source,ruby]
+----
+# bad
+super name, age
+
+# good
+super(name, age)
+----
+
+IMPORTANT: When calling `super` without arguments, `super` and `super()` mean different things. Decide what is appropriate for your usage.
+
 === Optional Arguments [[optional-arguments]]
 
 Define optional arguments at the end of the list of arguments.


README.adoc
```diff
@@ -2442,6 +2442,22 @@ puts temperance.age
 system 'ls'
 ----
 
+==== "super" method usage [[super-parens]]
+
+Always use parenthesis when calling `super` with arguments (keyword or not):
```
bbatsov: parentheses

README.adoc
```diff
@@ -2442,6 +2442,22 @@ puts temperance.age
...
+
+[source,ruby]
+----
+# bad
+super name, age
+super :name, :age
```
bbatsov: Why do symbols warrant a special case?

README.adoc
```diff
@@ -2442,6 +2442,22 @@ puts temperance.age
...
+# good
+super(name, age)
+super(:name, :age)
+----
+
+When calling `super` without arguments, `super` and `super()` mean different things. Decide what is appropriate for your usage.
```
bbatsov: I'd put `IMPORTANT:` in front of this (an admonition).

tejasbubane:
@bbatsov Fixed.

PR #756, opened 2019-06-12
Recommend style for float division coercion

### LINKED ISSUE #628: Introduce a recommended style for float division
By default, Ruby will perform integer division when both receiver and argument are integers. The way you perform float division on integers is to coerce one or more of them into a float. I propose there be a guideline for which part--receiver or argument--should take the `#to_f` coercion.

I am leaning towards both the receiver and the argument being coerced, because:

1. If you only coerce one, changing the code gets more risky, because you may unwittingly replace the float coerced one with an integer, regressing to integer division.

2. ~~Choosing one to apply `#to_f` to is (as far as I can see) completely arbitrary.~~

**Possible styles:**

```ruby
# both
a.to_f / b.to_f

# receiver
a.to_f / b

# argument
a / b.to_f
```

Another option could be to recommend `#fdiv`:

```ruby
a.fdiv(b)
```

WDYT?

tejasbubane:
Closes #628.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -3768,6 +3768,20 @@ rand(6) + 1
 rand(1..6)
 ----
 
+=== Float Division [[float-division]]
+
+When performing float-division on two integers, either use `fdiv` or convert the left-side integer (receiver) to float.
+
+[source,ruby]
+----
+# bad
+a / b.to_f
+
+# good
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #757, opened 2019-06-13
Fix syntax highlighting for some snippets

tejasbubane:
Using `Ruby` does not highlight the snippet - need to use smallcase `ruby`.

eg:

Screenshot 2019-06-13 at 2 18 29 PM


koic:
Thanks!

PR #758, opened 2019-06-13
Fix misleading example usage of English library

### LINKED ISSUE #714: Misleading variable in the rule example about Perl-style variables
The `$LOAD_PATH` variable is available without requiring the `English` module, yet the following example leads the reader to believe otherwise. Please check the [documentation for the variables](http://ruby-doc.org/stdlib-2.4.0/libdoc/English/rdoc/English.html) that the module introduces and replace the variable in the example with a suitable one, e.g. `$$` and `$PROCESS_ID`.
```
# bad
$:.unshift File.dirname(__FILE__)

# good
require 'English'
$LOAD_PATH.unshift File.dirname(__FILE__)
```

tejasbubane:
Split examples into two parts - with and without `English`.

Closes #714.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4663,7 +4663,6 @@ end
 
 Avoid using Perl-style special variables (like `$:`, `$;`, etc).
 They are quite cryptic and their use in anything but one-liner scripts is discouraged.
-Use the human-friendly aliases provided by the `English` library.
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #760, opened 2019-06-14
Link to project in contributing section

abraham:
Where contributions should be made wasn't listed anywhere.


bbatsov:
It would be better to use a conditional directive to show different instructions on GitHub and in the HTML version, as referring to the repo while you're in the repo might seem a bit weird. See https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/conditional-preprocessor-directives.adoc

bbatsov:
Anyways, I'll tackle this myself at some point. Thanks!

PR #762, opened 2019-06-16
Fix SOLID wiki link

garettarrowood:
Before this adjustment, clicking on this link opened up this url: https://en.wikipedia.org/wiki/SOLID_/(object-oriented_design/)


koic:
Good catch! This looks good.

PR #766, opened 2019-06-20
Use `case` instead of `if-elsif`s

tejasbubane:
Closes https://github.com/rubocop-hq/ruby-style-guide/issues/764

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1172,6 +1172,32 @@ result = if some_condition; something else something_else end
 result = some_condition ? something : something_else
 ----
 
+=== `case` vs `if-else` [[case-vs-if-else]]
+
+Prefer `case` over `if-elsif` when compared value is same in each clause.
+
+[source,ruby]
+----
+# bad
+if status == :active
+  perform_action
+elsif status == :inactive || status == :hibernating
+  check_timeout
+else
+  final_action
+end
+
+# good
+case status
+when :active
+  perform_action
+when :inactive, :hibernating
+  check_timeout
+else
+  final_action
+end
+----
+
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1164,6 +1164,32 @@ result = if some_condition; something else something_else end
 result = some_condition ? something : something_else
 ----
 
+=== Use `case` [[use-case]]
+
+Prefer `case` over `if-elsif`.
```
pirj: Do you think it makes sense to mention that this only applies to the case when the value being compared is always the same?
I.e.
```
if foo == :foo
  foo_action
elsif bar == :bar
  bar_action
end
```
doesn't really apply, since can't be expressed nicely with `case`.

bbatsov: Agreed.

README.adoc
```diff
@@ -1164,6 +1164,32 @@ result = if some_condition; something else something_else end
 result = some_condition ? something : something_else
 ----
 
+=== Use `case` [[use-case]]
```
bbatsov: We should come up with a better name for the title and the anchor. Something like `case vs if/else` or `When to use case?`. Many of the headings and anchors are really bad today for historical reasons - titles didn't exist until recently and were auto-generated from anchors which admittedly could have been much better.

README.adoc
tejasbubane: Fixed. Improved the example as well.

README.adoc
tejasbubane: Changed.

bbatsov:
Thanks!

PR #775, opened 2019-09-16
Use StandartError over Exception

denys281:
Based on discussion: https://github.com/rubocop-hq/rubocop/pull/7325/files#r320964544 

Rubocop issue: https://github.com/rubocop-hq/rubocop/issues/7299

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1931,6 +1931,19 @@ end
 
 Prefer the use of exceptions from the standard library over introducing new exception classes.
 
+=== Do not `raise` `Exception` [[raise-exception]]
+
+Use `StandardError` over `Exception`.
+
+[source,ruby]
+----
+# bad
+raise Exception, 'Error message here'
+
+# good
+raise StandardError, 'Error message here'
+----
+
 == Assignment & Comparison
 
 === Parallel Assignment [[parallel-assignment]]


README.adoc
```diff
@@ -1931,6 +1931,19 @@ end
 
 Prefer the use of exceptions from the standard library over introducing new exception classes.
 
+=== Do not `raise` `Exception` [[raise-exception]]
+
+Use `StandardError` over `Exception`.
```
bquorning: Perhaps this section should include a sentence about _why_ one should not `raise Exception`.

denys281: Could you help with wording?

denys281:
@marcandre  Could you take a look? Cop w merged some time ago.

marcandre:
This PR still needs work...
It's true one should not raise an `Exception`, but it is not true one should raise a `StandardException`; it's more an abstract base class than anything else.
One should raise the appropriate exception (e.g. `TypeError`) or if nothing fits, a `RuntimeError` (which shouldn't be specified).
Maybe it would be best to modify the "don't rescue `Exception`" section, as they are related... Don't raise one because you should not rescue those...

pirj:
Ping @denys281

PR #776, opened 2019-09-20
Introduce a reference .editorconfig file

### LINKED ISSUE #774: Introduce a reference .editorconfig file
Why not (also) having a reference .editorconfig file added to the guide, for main rules? With such file, the IDEs may help the user enforcing the rules. Something like:

```
root = true
charset = utf-8
end_of_line = lf
indent_style = space
insert_final_newline = true
max_line_length = 72
tab_width = 8
trim_trailing_whitespace = true

[*.rb]
indent_size = 2

[*.md]
# Four spaces are defined as the unit of indentation.
indent_size = 4

[*.json]
indent_size = 4
```

or similar. See https://editorconfig.org/

marcoXbresciani:
This patch adds a simple EditorConfig file as reference for the Ruby style guide.

See https://editorconfig.org/ for more details: EditorConfig helps maintain consistent coding styles for multiple developers working on the same project across various editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles. EditorConfig files are easily readable and they work nicely with version control systems.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4859,6 +4859,57 @@ TIP: RuboCop's cops (code checks) have links to the guidelines that they are bas
 
 http://www.jetbrains.com/ruby/[RubyMine]'s code inspections are http://confluence.jetbrains.com/display/RUBYDEV/RubyMine+Inspections[partially based] on this guide.
 
+=== EditorConfig
+
+EditorConfig helps maintain consistent coding styles for multiple
+developers working on the same project across various editors and IDEs.
+The EditorConfig project consists of a file format for defining coding
+styles and a collection of text editor plugins that enable editors to
+read the file format and adhere to defined styles. EditorConfig files
+are easily readable and they work nicely with version control systems.
+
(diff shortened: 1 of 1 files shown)


pirj:
EditorConfig is awesome.

Unfortunately, a reference to it does't fit this section that says:

> Here are some tools to **help you automatically check Ruby code against this guide**

Even though it can be configured to check Ruby code against this guide, it can be configured to do otherwise. While e.g. RuboCop strives to align its defaults with the guide.

To my best knowledge (limited to Idea, Eclipse, NetBeans, Vim, TextMate and Sublime), editors come with sane defaults or are extended with language-specific plugins. Which e.g. support better file type by file name detection (see [this list](https://github.com/rubocop-hq/rubocop/blob/3ade2760855459e86eeb71e6584115d3c204339d/config/default.yml#L11) for an example of Ruby code source file name patterns), and go even further than EditorConfig.

PR #777, opened 2019-09-20
add 'prepend' to Consistent Classes section

Mifrill:
Add missed `prepend` keyword into the description of class definitions:
https://github.com/rubocop-hq/ruby-style-guide#consistent-classes

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2622,9 +2622,10 @@ Use a consistent structure in your class definitions.
 [source,ruby]
 ----
 class Person
-  # extend and include go first
+  # extend/include/prepend go first
   extend SomeModule
   include AnotherModule
+  prepend YetAnotherModule
 
   # inner classes
   CustomError = Class.new(StandardError)


Mifrill:
perhaps it should better to put it after `private` section, because of this:
https://github.com/Mifrill/practice/blob/f1b46d070a15e98f5bf8616cf474cdf7f9c60cbd/prepend.rb#L1-L49
is:
```
module A
  def method_4
    puts __method__
  end

  private

  def method_5
    puts __method__
  end
end

class Prepend
  def self.method_1
    puts __method__
  end

  def method_2
    puts __method__
  end

  private

  def method_3
    puts __method__
  end

  prepend A
end

Prepend.method_1
Prepend.new.method_2
begin
  Prepend.new.method_3
rescue Exception => e
  puts e
end
Prepend.new.method_4
begin
  Prepend.new.method_5
rescue Exception => e
  puts e
end

```
output:
```
# ruby prepend.rb
method_1
method_2
private method `method_3' called for #<Prepend:0x000055e152ddac50>
Did you mean?  method
               method_2
               method_4
               methods
method_4
private method `method_5' called for #<Prepend:0x000055e152ddac50>
Did you mean?  method
               method_2
               method_4
               methods
```

bbatsov:
I was under the impression most people use it around `include/extend`. Haven't check actual code usages, though.

pirj:
In practice, I see no difference between `include` and `prepend` if they add a method that doesn't exist in the including class, does not matter if they are below `private` in the including class or above it.

pirj:
Moved `prepend` up next to `extend`/`include`.

Mifrill:
@pirj well, the idea was to highlight the behavior of `prepend` about `add method after the previous declarations:

```
module Included
  def included
    puts __method__
  end
end

module Prepended
  def prepended
    puts __method__
  end
end

class MyClass
  include Included

  def included
    puts "method redefined"
  end

  prepend Prepended

  def prepended
    puts "method redefined"
  end
end

puts MyClass.new.included
puts MyClass.new.prepended
```
output:
```
method redefined
prepended
```
so, if "standard" class structure like this:
```
class MyClass
 include MyInclude
 def method_1; end
 private
 def method_2; end
end
```
then the "reading" logic like: 
`define class -> define method (by include) -> redefine method (by class declaration)`

if we put `prepend` in top of class declaration:
```
class MyClass
 prepend MyInclude
 def method_1; end
 private
 def method_2; end
end
```
then the "reading" logic like: 
`define class -> redefine method (by prepend) -> define method (by class declaration)`

see? that's confusing...

So, to keep this "reading" logic, we should use the same stricture for `prepend`
```
class MyClass
 include MyInclude
 def method_1; end
 private
 def method_2; end
 prepend MyInclude
end
```
`define class -> define method (by include) -> redefine method (by class declaration)`
`define class -> define method (by class declaration) -> redefine method (by prepend)`

pirj:
To me, it would be confusing to have to scroll down to the bottom of the class, below private methods to find out that a certain class has features of a certain module.
As for the reader, `prepend` itself hints to me that the module overlays some methods that may (some of them may not!) be defined by the class.

So, first of all, I'd love to see **what** is included in the class.
And after that, probably - how it enhances the class.

Mifrill:
@pirj yeah, it makes sense too for me, well, shall I say: this point was raised for me someday, and I meet both definitions, so, that's why I've suggested clarifying this point with the community by style-guide.
Alright, we have 2 points here:
- the keyword `prepend` is missed in the description of class definitions:
https://github.com/rubocop-hq/ruby-style-guide#consistent-classes
(so, let's add this one)
- where and why we should put this keyword (top/bottom)
(so, let's decide it)

pirj:
Absolutely agree 👍

README.adoc
```diff
@@ -2622,9 +2622,10 @@ Use a consistent structure in your class definitions.
 [source,ruby]
 ----
 class Person
-  # extend and include go first
+  # extend/include/prepend go first
   extend SomeModule
   include AnotherModule
+  prepend MoreAnotherModule
```
andyw8, line 2628:
```suggestion
  prepend YetAnotherModule
```

pirj (approved):
Thank you!

PR #782, opened 2019-12-03
[Fix #763] Add rule suggesting not to use RUBY_VERSION in gemspec

### LINKED ISSUE #763: Do not use RUBY_VERSION in gemspec
Taken from https://github.com/rubocop-hq/rubocop/issues/7137

Gemspec should not contain `RUBY_VERSION` as a condition to switch dependencies.
For example:

```ruby
Gem::Specification.new do |s|
  if RUBY_VERSION >= '2.5'
    s.add_runtime_dependency 'gem_a'
  else
    s.add_runtime_dependency 'gem_b'
  end
end
```

It has unexpected behavior. The RUBY_VERSION is determined by `rake release`.
If I use Ruby 2.5 to execute `rake release`, the gem always has gem_a as a dependency even if a user installs the gem with Ruby 2.4 or older.

We have several ways to fix the problem.

* Remove the dependencies and use `post_install_message` to notice the gems.
* Add both gems to the dependency.
* If the dependencies are for development, move them to Gemfile.

@pocke Feel free to edit this.

tejasbubane:
Closes https://github.com/rubocop-hq/ruby-style-guide/issues/763

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4644,6 +4644,31 @@ https://ruby-doc.com/docs/ProgrammingRuby/html/rdtool.html[here].
 RD predated the rise of RDoc and YARD and was effectively obsoleted by them.footnote:[According to this https://en.wikipedia.org/wiki/Ruby_Document_format[Wikipedia article] the format used to be popular until the early 2000s when it was superseded by RDoc.]
 ****
 
+== Gemfile and Gemspec
+
+=== No `RUBY_VERSION` in the gemspec [[no-ruby-version-gemspec]]
+
+The gemspec should not contain `RUBY_VERSION` as a condition to switch dependencies.
+`RUBY_VERSION` is determined by `rake release`, so users may end up with wrong dependency.
+
+[source,ruby]
+----
+# bad
+Gem::Specification.new do |s|
+  if RUBY_VERSION >= '2.5'
+    s.add_runtime_dependency 'gem_a'
+  else
+    s.add_runtime_dependency 'gem_b'
+  end
+end
+----
+
+Fix by either:
+
+* Post-install messages.
+* Add both gems as dependency (if permissible).
+* If development dependencies, move to Gemfile.
+
 == Misc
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -4644,6 +4644,31 @@ https://ruby-doc.com/docs/ProgrammingRuby/html/rdtool.html[here].
 RD predated the rise of RDoc and YARD and was effectively obsoleted by them.footnote:[According to this https://en.wikipedia.org/wiki/Ruby_Document_format[Wikipedia article] the format used to be popular until the early 2000s when it was superseded by RDoc.]
 ****
 
+== Gemfile & Gemspec
```
koic: Can you use `and` instead of `&`?

```diff
-== Gemfile & Gemspec
+== Gemfile and Gemspec
```

README.adoc
```diff
@@ -4644,6 +4644,31 @@ https://ruby-doc.com/docs/ProgrammingRuby/html/rdtool.html[here].
...
+
+=== No RUBY_VERSION in gemspec [[no-ruby-version-gemspec]]
```
koic: Likely it is better to add `the` and backquotes.

```diff
-=== No RUBY_VERSION in gemspec [[no-ruby-version-gemspec]]
+=== No `RUBY_VERSION` in the gemspec [[no-ruby-version-gemspec]]
```

README.adoc
```diff
@@ -4644,6 +4644,31 @@ https://ruby-doc.com/docs/ProgrammingRuby/html/rdtool.html[here].
...
+
+Gemspec should not contain `RUBY_VERSION` as a condition to switch dependencies.
```
koic: Ditto.

```diff
-Gemspec should not contain `RUBY_VERSION` as a condition to switch dependencies.
+The gemspec should not contain `RUBY_VERSION` as a condition to switch dependencies.
```

tejasbubane:
@koic Fixed.

koic:
Thanks!

PR #783, opened 2019-12-05
Fix typo and minor formatting

### LINKED ISSUE #773: Typo in `Syntaces`
Syntaces should be Syntaxes. See https://rubystyle.guide/#no-mixed-hash-syntaces.

chrisjwelly:
Fixes #773.

Also, currently `if`s in "No semicolon `if`s" do not render the code block for `if`. I removed the 's' in order to allow it.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -3599,7 +3599,7 @@ hash = { :one => 1, :two => 2, :three => 3 }
 hash = { one: 1, two: 2, three: 3 }
 ----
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Btw, those two headings should be reworded to something like `Do not ...`, as the current weird versions were simply generated from the anchors when we changed the structure of the guide. We should also check if RuboCop links to those anchors before updating them here.

bbatsov:
Seems this also needs to be rebased.

marcandre:
Merged, thanks!

PR #785, opened 2019-12-31
Add rule stating control flow keywords should begin their line

Drenmi:
### Rationale

Breaking control flow is almost certainly the most significant thing that is going on in that line of code, and so it deserves the most prominent spot, at the beginning of the line.

Hiding away control flow in the middle of a long line makes it unnecessarily hard to parse, and Ruby gives us a slew of forms in which we can rewrite the line with the change in flow emphasized.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1052,6 +1052,19 @@ end
 
 == Flow of Control
 
+=== Control Flow Keywords [[control-flow-keywords]]
+
+Keywords that alter control flow, e.g. `next`, `raise`, `return`, etc. should appear at the beginning of the line.
+
+[source,ruby]
+----
+# bad
+makes_sense? ? "Great success!" : raise("Woah.")
+
+# good
+raise("Woah.") unless makes_sense?
+----
+
 === No `for` Loops [[no-for-loops]]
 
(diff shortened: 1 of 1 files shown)


koic:
The following idioms exist for `return` and` raise` in Ruby.

- `do_something and return`
- `do_something || raise`

For example, Rails framework will show users an error message using this idiom.

> `"redirect_to(...) and return\"`

https://github.com/rails/rails/blob/v6.0.2.1/actionpack/lib/abstract_controller/rendering.rb#L10

So I think these cases need to be accepted.

bbatsov:
Agreed. I guess we can word this better and provide more examples.

Drenmi:
Hm. I would add a note about its usage in Rails, but I would be hesitant to recommend it in the style guide. It being used in the wild I think should not be seen as an indication that it's good. 😬 

I would further argue `and return` is a Rails idiom, not a Ruby one. (At least I've only ever seen it in Rails codebases.) It is generally also not needed. I've worked with Rails full time for the past 6 years, and never written `and return`. 🙂 

Will update the PR with the rationale for this rule, as well as notes on the usage of `and return`.

PR #786, opened 2020-01-04
Improve space exeptions

BurdetteLamar:


bbatsov:
Thanks!

README.adoc
```diff
@@ -231,7 +231,9 @@ a, b = 1, 2
 class FooError < StandardError; end
 ----
 
-There are a few exceptions. One is the exponent operator:
+There are a few exceptions:
+
+* Exponent operator:
```
koic: It is not a continuous list item because it contains example code. In this case, I think it's better to use headings instead of list items.

```diff
-There are a few exceptions:
-
-* Exponent operator:
+There are a few exceptions.
+
+==== Exponent operator
```

BurdetteLamar: In that case, a level-4 header for 'Exceptions', and level-5 headers for the three exceptions?

Would all that show up in the table of conents?

PR #789, opened 2020-01-15
Useless block arg comma

pawptart:
In reference to https://github.com/rubocop-hq/rubocop/pull/7637.

Adds style guide for block arguments to omit any trailing commas except in single argument cases where functionality would be affected.


bbatsov:
You might some clarification that when you have a single param this is actually doing destructuring and that's the difference.

README.adoc
```diff
@@ -2194,6 +2194,32 @@ with_tmp_dir do |dir|
...
+  a + b + c
+end
+
+# good
+[[1, 2, 3], [4, 5, 6]].each do |a, b, c|
+  a + b + c
+end
+
+# bad
+[[1, 2, 3], [4, 5, 6]].each { |a, b, c| a + b + c }
```
dadah89: this bad example is the same as the good one just below

pawptart: Fixed.

bbatsov:
Thanks!

PR #790, opened 2020-01-26
Recommend Hash#transform_keys & transform_values

djudd-stripe:
These methods are more readable than the more flexible alternatives when
only keys or only values are being transformed, and also execute more
quickly in microbenchmarks.

Corresponding Rubocop PR here: https://github.com/rubocop-hq/rubocop/pull/7663


bbatsov:
Thanks!

PR #791, opened 2020-02-14
Fix Explicit `self` example

jibiel:
I believe that `self` was avoided too much in the example and its previous version contradicted the rule description above.


koic:
#780 merge was incorrect. Thanks!

PR #807, opened 2020-03-19
Clarify meaning of methods with "keyword" status

### LINKED ISSUE #559: Methods that have "keyword" status?
In the rule that governs the use of parentheses in method invocations, the part

> Omit parentheses around parameters for ... methods that have "keyword" status in Ruby (e.g. `attr_reader`, `puts`)

isn't clear, especially for people new to Ruby, as to what exactly is "keyword" status in Ruby, or which methods.

I think we should elaborate on it to make it instructive enough to be useful : )

jlmuir:
Clarify the meaning of methods that have so-called "keyword" status.

Expand the section on the omission or use of parentheses in calls of
non-declarative methods with "keyword" status by explaining the two
schools of thought accepted by the community.

Fixes #559.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2525,6 +2525,14 @@ validates :name, presence: true
 
 Always omit parentheses for methods that have "keyword" status in Ruby.
 
+NOTE: Unfortunately, it's not exactly clear _which_ methods have "keyword" status.
+There is agreement that declarative methods have "keyword" status.
+However, there's less agreement on which non-declarative methods, if any, have "keyword" status.
+
+===== Declarative Methods That Have "keyword" Status in Ruby [[method-invocation-parens-declarative-keyword]]
+
+Always omit parentheses for declarative methods (a.k.a. DSL methods or macro methods) that have "keyword" status in Ruby (e.g., various `Module` instance methods):
+
 [source,ruby]
 ----
 class Person
@@ -2537,18 +2545,26 @@ class Person
 end
 ----
 
-==== Non-declarative Methods That Have "keyword" Status in Ruby [[method-invocation-parens-non-declarative-keyword]]
+===== Non-Declarative Methods That Have "keyword" Status in Ruby [[method-invocation-parens-non-declarative-keyword]]
 
-Can omit parentheses for methods that have "keyword" status in Ruby, but are not declarative:
+For non-declarative methods with "keyword" status (e.g., various `Kernel` instance methods), two styles are considered acceptable.
+By far the most popular style is to omit parentheses.
+Rationale: The code reads better, and method calls look more like keywords.
+A less-popular style, but still acceptable, is to include parentheses.
+Rationale: The methods have ordinary semantics, so why treat them differently, and it's easier to achieve a uniform style by not worrying about which methods have "keyword" status.
+Whichever one you pick, apply it consistently.
 
 [source,ruby]
 ----
-# good
-puts(temperance.age)
-system('ls')
-# also good
+# good (most popular)
 puts temperance.age
 system 'ls'
+exit 1
+
+# also good (less popular)
+puts(temperance.age)
+system('ls')
+exit(1)
 ----
 
 ==== Using `super` with Arguments  [[super-with-args]]


README.adoc
```diff
@@ -2523,7 +2523,14 @@ validates :name, presence: true
 
 ==== Methods That Have "keyword" Status in Ruby [[method-invocation-parens-keyword]]
 
-Always omit parentheses for methods that have "keyword" status in Ruby.
+The set of methods that have "keyword" status is not well-defined.
```
bbatsov: I think we still have to provide a guideline here (even if we soften the language) and add the extra information in some admonition block (e.g. NOTE:).

README.adoc
```diff
@@ -2537,18 +2544,29 @@ class Person
 end
 ----
 
-==== Non-declarative Methods That Have "keyword" Status in Ruby [[method-invocation-parens-non-declarative-keyword]]
+===== Non-Declarative Methods That Have "keyword" Status in Ruby [[method-invocation-parens-non-declarative-keyword]]
+
+There are two schools of thought within the community: one says to always omit parentheses for non-declarative methods that have "keyword" status (e.g., `Kernel` instance methods), and the other says not to omit them.
```
bbatsov: The are two options indeed, but one is much more popular, so I guess we'd still have to recommend it and just mention the other.

README.adoc
```diff
@@ -2537,18 +2544,29 @@ class Person
...
 
-Can omit parentheses for methods that have "keyword" status in Ruby, but are not declarative:
+The omit-parentheses school says that certain non-declarative methods feel like they have "keyword" status.
+Since they feel like keywords, the code reads better *without* parentheses because it looks like keywords rather than method calls.
+
+The use-parentheses school says that there's nothing special about these non-declarative methods, their semantics are ordinary, and which methods have "keyword" status is not well-defined.
+Therefore, *with* parentheses, it's easier to write with a consistent style since you don't have to worry about defining which non-declarative methods have "keyword" status, and the code reads better because it's just ordinary method calls.
+
+There are respected Rubyists from both schools of thought within the community, so choose the one you identify with most and be consistent.
```
bbatsov: As this can be repeated about several guidelines I'd rather skip it.

README.adoc
```diff
@@ -2537,18 +2544,29 @@ class Person
...
-==== Non-declarative Methods That Have "keyword" Status in Ruby [[method-invocation-parens-non-declarative-keyword]]
+===== Non-Declarative Methods That Have "keyword" Status in Ruby [[method-invocation-parens-non-declarative-keyword]]
+
+There are two schools of thought within the community: one says to always omit parentheses for non-declarative methods that have "keyword" status (e.g., `Kernel` instance methods), and the other says not to omit them.
 
-Can omit parentheses for methods that have "keyword" status in Ruby, but are not declarative:
+The omit-parentheses school says that certain non-declarative methods feel like they have "keyword" status.
+Since they feel like keywords, the code reads better *without* parentheses because it looks like keywords rather than method calls.
+
+The use-parentheses school says that there's nothing special about these non-declarative methods, their semantics are ordinary, and which methods have "keyword" status is not well-defined.
```
bbatsov: Probably I'd stick to the more common style (omitting parens) and mention the alternative in a nice callout block.

README.adoc
```diff
@@ -2523,7 +2523,14 @@ validates :name, presence: true
 
 ==== Methods That Have "keyword" Status in Ruby [[method-invocation-parens-keyword]]
 
-Always omit parentheses for methods that have "keyword" status in Ruby.
+The set of methods that have "keyword" status is not well-defined.
+There is agreement that declarative methods have "keyword" status, but which methods are declarative methods is not well-defined.
```
bbatsov: I'm not sure how common the term "declarative" is in Ruby. More often than not I've seen those called DSL methods or macro methods. Might make sense to mention all terms for clarity.

showyi81929:
The use-parentheses school says that there's nothing special about these non-declarative methods, their semantics are ordinary, and which methods have "keyword" status is not well-defined.
Therefore, *with* parentheses, it's easier to write with a consistent style since you don't have to worry about defining which non-declarative methods have "keyword" status, and the code reads better because it's just ordinary method calls.

 There are respected Rubyists from both schools of thought within the community, so choose the one you identify with most and be consistent.

README.adoc
jlmuir: OK, I force-pushed an amended commit that provides a guideline and adds extra info in an admonition block.

README.adoc
jlmuir: OK, I force-pushed an amended commit that makes it clear that one is much more popular.

README.adoc
jlmuir: OK, I force-pushed an amended commit that removes that.

README.adoc
jlmuir: OK, I force-pushed an amended commit that doesn't go so far as to stick with the omitting-parens style and only mentioning the alternative in a callout block, but it does make it clear that one is much more popular than the other.

README.adoc
jlmuir: OK, I force-pushed an amended commit that adds the other terms for clarity.

PR #808, opened 2020-03-20
Refine an example for "Suppressing Exceptions" rule

koic:
Follow up of https://github.com/rubocop-hq/rubocop/pull/7805#issuecomment-601071106.

This PR refines an example for "Suppressing Exceptions" rule.

And `do_something rescue nil` will be changed from bad case to good case based on RuboCop (0.81)'s behavior and the comment below.
https://github.com/rubocop-hq/rubocop/pull/7297#issuecomment-523501395

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1813,12 +1813,25 @@ Don't suppress exceptions.
 ----
 # bad
 begin
-  # an exception occurs here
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #814, opened 2020-04-26
Refine an example for "Space in Method Calls" rule

koic:
Follow up of https://github.com/rubocop-hq/rubocop/pull/7909.

This PR refines an example for "Space in Method Calls" rule.

The example code is not compatible.

```ruby
def f(num)
  num * 2
end

p f (3 + 2) + 1 #=> 12
p f(3 + 2) + 1  #=> 11
```

This PR replaces an example code.


bbatsov:
Thanks!

PR #815, opened 2020-04-28
Use the term CapitalCase instead of CamelCase

rpbarlow-credible:
The document referred to what is known as CapitalCase as
CamelCase. camelCase means that the first letter is lowercase,
while subsequent words are capital letters. CapitalCase is when
all words start with a capital.

Signed-off-by: Randy Barlow <rbarlow@credible.com>

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -834,9 +834,9 @@ def some_method1
 end
 ----
 
-=== CamelCase for Classes [[camelcase-classes]]
+=== CapitalCase for Classes [[camelcase-classes]]
 
-Use `CamelCase` for classes and modules.
+Use `CapitalCase` for classes and modules.
 (Keep acronyms like HTTP, RFC, XML uppercase).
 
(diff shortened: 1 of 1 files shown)


bbatsov:
I think we already had a similar suggestion about the term `PascalCase`. Probably we can add some footnotes or a sidebar on the subject, but I'm pretty sure most people can't tell the `CamelCase` from `CapitalCase`.

marcandre:
Wikipedia has [a few variations](https://en.wikipedia.org/wiki/Camel_case#Variations_and_synonyms) and use "upper camel case" internally.

I feel that using `CapitalCase`, `CapitalWords` as in Python, or`PascalCase`  are all superior to `CamelCase` because they won't introduce any cognitive dissonance.

bbatsov:
No argument from me. I just think it'd be nice if we expanded somewhere on this, as different people grew up referring to this differently.

PR #817, opened 2020-04-30
Add new "Empty Lines around Attribute Accessor" rule

koic:
Follow up to #816.
This PR adds new "Empty Lines around Attribute Accessor" rule. It will update the existing example with the new rule name.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -478,9 +478,9 @@ some_method
 some_method
 ----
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

bbatsov:
Technically speaking there should be an empty line between the accessors and whatever follows them, but this can be improved latter. The rule can be made more generic about grouping related class macro-style methods.

PR #819, opened 2020-05-17
Use `filter_map` instead of find_all/select/filter + map

tejasbubane:
Rubocop Issue: https://github.com/rubocop-hq/rubocop/issues/7946


bbatsov:
I wouldn't say this reads better, so I'm wondering if that's a style or a performance consideration. From what I gathered the reason for the new method is speed.

tejasbubane:
On second thought, I agree that this does not read better. 

The cop should most probably be added to [rubocop-performance](https://github.com/rubocop-hq/rubocop-performance).

PR #822, opened 2020-06-30
More clear (?) example for Exponential Notation

ghiculescu:
The current formatting for this section looks broken:

image
image

Rather than mess around with ** (which is being misread as markdown), I just spelt it out using the wording they taught me in high school math. Hopefully I remember it right...

| power notation | exponential notation | output   |
|----------------|----------------------|----------|
| 10 ** 7        | 1e7                  | 10000000 |
| 10 ** 6        | 1e6                  | 1000000  |
| 10 ** 7        | 10e6                 | 10000000 |

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4064,6 +4064,12 @@ When using exponential notation for numbers, prefer using the normalized scienti
 
 The goal is to avoid confusion between powers of ten and exponential notation, as one quickly reading `10e7` could think it's 10 to the power of 7 (one then 7 zeroes) when it's actually 10 to the power of 8 (one then 8 zeroes). If you want 10 to the power of 7, you should do `1e7`.
 
+| power notation | exponential notation | output   |
+|----------------|----------------------|----------|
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -4062,7 +4062,7 @@ a.fdiv(b)
 
 When using exponential notation for numbers, prefer using the normalized scientific notation, which uses a mantissa between 1 (inclusive) and 10 (exclusive). Omit the exponent altogether if it is zero.
 
-The goal is to avoid confusion between powers of ten and exponential notation, as one quickly reading `10e7` could think it's `10**7` when it's actually `10**8`.
```
bbatsov, line 4065: I think a simpler fix would be to use the passthrough notation (`+ .. +`). See https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#formatted-text

bbatsov: Btw, AsciiDoc actually has superscripts, so I guess that'd be even better - `10^7^`.

PR #830, opened 2020-07-11
From Perl's POD to RD - :begin & :end formatting

MSP-Greg:
Small formatting change

Closes https://github.com/rubocop-hq/ruby-style-guide/issues/829


bbatsov:
Thanks!

PR #833, opened 2020-08-01
Add "Float Comparison" section

fatkodima:


marcandre:
LGTM. Did you want to mention the only exception I can think of which is `if x == Float::INFINITY`?

fatkodima:
Updated. Also mentioned `Float::NAN`.

marcandre:
Actually you should never compare to `NAN`:
```ruby
Float::NAN == Float::NAN # => false!!
```

fatkodima:
Yes, sure, my bad, you're right, sorry. Removed that.

bbatsov:
We need a lint cop for that! :D

PR #834, opened 2020-08-05
Add comma to Multi-line Method Chains sentence

jlmuir:
In the Multi-line Method Chains section, the Leading and Trailing style
subsections both begin with a similar sentence, yet the sentence in the
Leading subsection does not use a comma after the introductory clause,
while the sentence in the Trailing subsection does.  Add a comma to the
Leading subsection sentence to make the two sentences consistent with
each other.


marcandre:
Thanks!

PR #835, opened 2020-08-08
Add new "Identity Comparison" rule

koic:
This PR adds new "Equal Object Identifier" rule that prefers `equal?` over `==` when comparing `object_id`. `Object#equal?` method is provided to compare objects for identity, and comparing objects for equality using `Object#==` does not require
redundant `object_id` to be repeated.

```ruby
# bad
foo.object_id == bar.object_id

# good
foo.equal?(bar)
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2198,6 +2198,19 @@ something = something && something.downcase
 something &&= something.downcase
 ----
 
+=== Identity Comparison [[identity-comparison]]
+
+Prefer `equal?` over `==` when comparing `object_id`. `Object#equal?` is provided to compare objects for identity, and in contrast `Object#==` is provided for the purpose of doing value comparison.
+
+[source,ruby]
+----
+# bad
+foo.object_id == bar.object_id
+
+# good
+foo.equal?(bar)
+----
+
 === Explicit Use of the Case Equality Operator [[no-case-equality]]
 
 Avoid explicit use of the case equality operator `===`.


README.adoc
```diff
@@ -2198,6 +2198,19 @@ something = something && something.downcase
 something &&= something.downcase
 ----
 
+=== Equal Object Identifier [[equal-object-id]]
```
bbatsov, line 2201: I think we can name this "Identity Comparison", as the other form of comparison is "Value Comparison" (the one normally done with `==`).

README.adoc
```diff
@@ -2198,6 +2198,19 @@ something = something && something.downcase
...
+
+Prefer `equal?` over `==` when comparing `object_id`. `Object#equal?` method is provided to compare objects for identity, and comparing objects for equality using `Object#==` does not require redundant `object_id` to be repeated.
```
bbatsov, line 2203: > comparing objects for equality using `Object#==` does not require redundant `object_id` to be repeated.

I think this part can be worded better, or changed to something like "and in contrast `Object#==` is provided for the purpose of doing value comparison" or something along those lines.

koic:
Thank you for your review and sophisticated suggestions!  I updated this PR.

bbatsov:
You're welcome!

marcandre:
Created #836 as a followup...

PR #849, opened 2020-10-25
Remove architecture entries

andyw8:
These are applicable to any object oriented language, and are about higher level design concepts rather than style. I think they should be removed to keep the style guide focused.


bbatsov:
Perhaps, altough I think that adding some examples to such guidelines actually makes them useful.

PR #850, opened 2020-10-25
Fix grammar for 'No Extend'

andyw8:


README.adoc
```diff
@@ -3239,7 +3239,7 @@ Person = Struct.new(:first_name, :last_name) do
 end
 ----
 
-=== No Extend `Struct.new` [[no-extend-struct-new]]
+=== Don't Extend `Struct.new` [[dont-extend-struct-new]]
```
bbatsov, line 3242: Please, check that there's no reference to this in RuboCop, as the anchor might need to be updated there as well.

andyw8: There was indeed, so I've changed it back.

bbatsov:
Eventually we'll have to bite the bullet and update the anchors as well, but it's not that important.

PR #851, opened 2020-10-25
Remove entry for 'Comment Upkeep'

andyw8:
This would apply to any programming language.


bbatsov:
Many entries in the style guide are generic, so I'd rather keep it.

PR #854, opened 2020-11-26
Added note and example about precedence of and/or operator.

bipashant:
Added note and example about the precedence of and/or operator.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1368,12 +1368,16 @@ The `and` and `or` keywords are banned.
 The minimal added readability is just not worth the high probability of introducing subtle bugs.
 For boolean expressions, always use `&&` and `||` instead.
 For flow control, use `if` and `unless`; `&&` and `||` are also acceptable but less clear.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1368,13 +1368,17 @@ The `and` and `or` keywords are banned.
 The minimal added readability is just not worth the high probability of introducing subtle bugs.
 For boolean expressions, always use `&&` and `||` instead.
 For flow control, use `if` and `unless`; `&&` and `||` are also acceptable but less clear.
+Please note that the `and` and `or` operators have lower precedence than the `=` operator, whereas the `&&` and `||` operators have higher precedence than the `=` operator, based on order of operations.
```
bbatsov, line 1371: I'd probably mention here also that `&&` has higher precedence than `||`, but `and` and `or` have the same one, which IMO is another source of confusion and potential bugs.

bipashant: @bbatsov 
Good point. I've updated the pull request.

bbatsov:
Thanks!

PR #855, opened 2020-11-29
Fix Rubocop plugins broken link

gabrielsenaco:
Changes the [broken link](https://docs.rubocop.org/en/stable/integration_with_other_tools/) on RuboCop section to [updated link](https://docs.rubocop.org/rubocop/integration_with_other_tools.html).

Before the adjustment, clicking on https://docs.rubocop.org/en/stable/integration_with_other_tools/ leads to an invalid page.


bbatsov:
Thanks!

PR #856, opened 2020-12-15
Add new "Space in Brackets Access" rule

koic:
This PR adds "Space in Brackets Access" rule.

```ruby
# bad
collection [index_or_key]

# good
collection[index_or_key]
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -786,6 +786,19 @@ puts (x + y)
 puts(x + y)
 ----
 
+=== Space in Brackets Access
+
+Do not put a space between a receiver name and the opening brackets.
+
+[source,ruby]
+----
+# bad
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -786,6 +786,19 @@ puts (x + y)
 puts(x + y)
 ----
 
+=== Space in Brackets Access [[brackets-no-spaces]]
```
bbatsov, line 789: There's no need for the explicit anchor,  as one will be generated anyways (and it seems to me it'd be more descriptive). All the short unclear anchors are legacy from a time we didn't think enough about them. :-)

koic: I get it! That makes sense!

PR #861, opened 2021-01-30
Add rule for using `to_h` instead of `Hash[]`

tejasbubane:
Issue: https://github.com/rubocop-hq/rubocop/issues/9460

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4002,6 +4002,25 @@ When you've got keys that are not symbols stick to the hash rockets syntax.
 { :a => 1, 'b' => 2 }
 ----
 
+=== Avoid Hash[] constructor [[avoid-hash-constructor]]
+
+`Hash::[]` was pre-Ruby 2.1 way of constructing hashes from arrays of key-value pairs,
+or from a flat list of keys and values. It has an obscure semantic and looks cryptic in code.
+Since Ruby 2.1, `Enumerable#to_h` can be used to construct a hash from a list of key-value pairs,
+and it should be preferred. Instead of `Hash[]` with a list of literal keys and values,
+just a hash literal should be preferred.
+
+[source,ruby]
+----
+# bad
+Hash[ary]
+Hash[a, b, c, d]
+
+# good
+ary.to_h
+{a => b, c => d}
+----
+
 === `Hash#key?` [[hash-key]]
 
 Use `Hash#key?` instead of `Hash#has_key?` and `Hash#value?` instead of `Hash#has_value?`.


bbatsov:
//cc @zverok

README.adoc
```diff
@@ -4000,6 +4000,23 @@ When you've got keys that are not symbols stick to the hash rockets syntax.
 { :a => 1, 'b' => 2 }
 ----
 
+=== Array to hash [[array-to-hash]]
```
zverok, line 4003: For the same reason we discussed in Rubocop's PR, I believe "array" shouldn't be here :)
Something like "Don't use `Hash[]` constructor", or something.

tejasbubane: changed @zverok

README.adoc
```diff
@@ -4000,6 +4000,23 @@ When you've got keys that are not symbols stick to the hash rockets syntax.
 { :a => 1, 'b' => 2 }
 ----
 
+=== to_h over Hash[] [[to-h-over-hash]]
```
zverok, line 4003: Also not it :( Of two "bad→good" examples, one doesn't use (and doesn't need) `#to_h` at all.
I really believe it should say something along the lines

> **Avoid `Hash[]` constructor**
>
> `Hash::[]` was pre-Ruby 2.1 way of constructing hashes from arrays of key-value pairs, or from a flat list of keys and values. It has an obscure semantic and looks cryptic in code. Since Ruby 2.1, `Enumerable#to_h` can be used to construct a hash from a list of key-value pairs, and it should be preferred. Instead of `Hash[]` with a list of literal keys and values, just a hash literal should be preferred.

Or something like that.

tejasbubane: This is much better :+1: changed.

tejasbubane:
@zverok @bbatsov Can we get this merged please?

README.adoc
```diff
@@ -4002,6 +4002,25 @@ When you've got keys that are not symbols stick to the hash rockets syntax.
...
+=== Avoid Hash[] constructor [[avoid-hash-constructor]]
+
+`Hash::[]` was pre-Ruby 2.1 way of constructing hashes from arrays of key-value pairs,
+or from a flat list of keys and values. It has an obscure semantic and looks cryptic in code.
+Since Ruby 2.1, `Enumerable#to_h` can be used to construct a hash from a list of key-value pairs,
+and it should be preferred. Instead of `Hash[]` with a list of literal keys and values,
+just a hash literal should be preferred.
+
+[source,ruby]
+---
```
bbatsov, line 4014: You're missing a `-` here. (they have to be 4)

README.adoc
```diff
@@ -4002,6 +4002,25 @@ When you've got keys that are not symbols stick to the hash rockets syntax.
...
+# bad
+Hash[ary]
+Hash[a, b, c, d]
+
+# good
+ary.to_h
+{a => b, c => d}
+---
```
bbatsov, line 4022: Same here.

README.adoc
tejasbubane: Fixed.

README.adoc
tejasbubane: Fixed.

PR #863, opened 2021-01-31
[Fixes #613] Avoid tap

### LINKED ISSUE #613: Adding tap method usage
[Object#tap](http://apidock.com/ruby/Object/tap) is very useful in rare situations and makes code more readable. Can we add a section related to guidlines ragarding good usage of tap method? I can create a pull request if someone can guide me

marcandre:


bbatsov:
I doubt that anyone can argue with this suggestion. Thanks!

PR #866, opened 2021-02-22
Introduce `unless` with Multiple Conditions guideline

pirj:
Boolean logic with negations is hard.

Fixes #623

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1527,6 +1527,34 @@ else
 end
 ----
 
+=== `unless` with Multiple Conditions
+
+Do not use `unless` with multiple conditions.
+
+[source,ruby]
+----
+# bad
+unless sweet?(fruit) && !fruit.green?
+  yield fruit # skip yielding sweet non-green fruits
+end
+
+# bad
+return unless node.ternary? && !infinite_loop? && offense?(node)
+
+# good - non-inversed
+if !sweet?(fruit) || fruit.green?
+  yield fruit # yield non-sweet or green fruits
+end
+
+# good - all separate
+return unless node.ternary?
+return if infinite_loop?
+return unless offense?(node)
+
+# good - explicit intent
+return unless offensive_ternary?(node)
+----
+
 === Parentheses around Condition [[no-parens-around-condition]]
 
 Don't use parentheses around the condition of a control expression.


README.adoc
```diff
@@ -1527,6 +1527,34 @@ else
...
+=== `unless` with Multiple Conditions
+
+Do not use `unless` with multiple conditions.
+
+[source,ruby]
+----
+# bad
+unless sweet?(fruit) && !fruit.green?
+  yield fruit # skip yielding sweet non-green fruits
+end
```
pirj, line 1539: This is taken from the same source code as an example below 😉, just changed the names :spy:

README.adoc
```diff
@@ -1527,6 +1527,34 @@ else
...
 ----
 
+=== `unless` with Multiple Conditions
+
+Do not use `unless` with multiple conditions.
+
+[source,ruby]
+----
+# bad
+unless sweet?(fruit) && !fruit.green?
```
andyw8, line 1537: For me, the problem isn't so much the multiple conditions, if it's the mixing of `unless` with the inversion. I'd have no problem with `unless sweet?(fruit) && fruit.green?`.

Let's try get some wider opinions – but I think this could be a tricky area to reach a consensus on.

bbatsov:
There's also some related conversation here https://github.com/rubocop-hq/rubocop/issues/5388

While I agree we should discourage multiple conditions in `unless` probably we to distinguish between "okish" use-cases (e.g. `unless x && y` and more complex cases involving multiple logical ops and negation that are just hard to read.

README.adoc
bbatsov: Yeah, I agree we should have a separate listing for an "OKish" use-case, involving more straightforward cases.

pirj: I'm perfectly fine with narrowing this guideline down to:

> Do not use `unless` with multiple conditions that involve negation.

And adding example `unless sweet?(fruit) && fruit.green?` to OK'ish section.

Is this sufficient to convience you guys?

README.adoc
```diff
@@ -1527,6 +1527,34 @@ else
...
+  yield fruit # skip yielding sweet non-green fruits
+end
+
+# bad
+return unless node.ternary? && !infinite_loop? && offense?(node)
+
+# good - non-inversed
+if !sweet?(fruit) || fruit.green?
```
pirj, line 1545: In the same vein, I don't have an impression that `if` reads so well with one of the conditions negated. "If not sweet or green" sounds ambiguous. "If green or not sweet" is slightly better.

README.adoc
bbatsov: Well, I think things become dicey the moment you write something like `unless x && y || z`. My point was that a single `||` or `&&` without negation is ok for some people, but generally everything else is a bad idea in terms of readability.

pirj:
I may eventually make another attempt, but for now, closing as stalled.

PR #868, opened 2021-02-22
Prefer using "call" to "invoke" applied to methods

pirj:
I believe "invoke" is rarely used in the Ruby landscape, and "call" is used more often.

[`Method` has a `call` method](https://ruby-doc.org/core-3.0.0/Method.html#method-i-call).
[`Proc` has a `call` method](https://ruby-doc.org/core-3.0.0/Proc.html#method-i-call).

Also,

> As verbs the difference between call and invoke is that call is (lb) to use one's voice while invoke is to call upon (a person, especially a god) for help, assistance or guidance.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -672,7 +672,7 @@ There are two popular styles in the Ruby community, both of which are considered
 
 ==== Leading `.` [[leading-dot-in-multi-line-chains]]
 
-When continuing a chained method invocation on another line, keep the `.` on the second line.
+When continuing a chained method call on another line, keep the `.` on the second line.
 
 [source,ruby]
 ----
@@ -687,7 +687,7 @@ one.two.three
 
 ==== Trailing `.` [[trailing-dot-in-multi-line-chains]]
 
-When continuing a chained method invocation on another line, include the `.` on the first line to indicate that the expression continues.
+When continuing a chained method call on another line, include the `.` on the first line to indicate that the expression continues.
 
 [source,ruby]
 ----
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -4969,7 +4969,7 @@ Use `%r` only for regular expressions matching _at least_ one `/` character.
 
 === `%x` [[percent-x]]
 
-Avoid the use of `%x` unless you're going to invoke a command with backquotes in it (which is rather unlikely).
+Avoid the use of `%x` unless you're going to execute a command with backquotes in it (which is rather unlikely).
```
pirj, line 4972: https://ruby-doc.org/core-2.7.1/Kernel.html#method-i-system

> Executes command

README.adoc
```diff
@@ -2723,9 +2723,9 @@ def some_method_with_parameters(param1, param2)
 end
 ----
 
-=== Method Invocation Parentheses [[method-invocation-parens]]
+=== Method Call Parentheses [[method-invocation-parens]][[method-call-parens]]
```
pirj, line 2726: Kept old anchors.

AsciiDoc reference for additional anchors https://docs.asciidoctor.org/asciidoc/latest/sections/custom-ids/#assign-additional-ids

bbatsov:
Fine by me. I'm reasonably sure the usage of invocation came from me, as I've alwasy been fond of the term, but I'm certainly not attached to it. :-)

bbatsov:
> As verbs the difference between call and invoke is that call is (lb) to use one's voice while invoke is to call upon (a person, especially a god) for help, assistance or guidance.

Btw, this actually confirms that "invoke" is the better term in the context of programming. :D

bbatsov:
Here there's a nice conversation on the subject - https://stackoverflow.com/questions/50884893/calling-vs-invoking-a-function

PR #873, opened 2021-05-28
Update "`then` in Multi-line Expression" rule

koic:
This PR adds Ruby 2.7's `in` pattern syntax to "`then` in Multi-line Expression" rule.


pirj (approved):
👍 Thanks

PR #878, opened 2021-06-18
feat: flip-flop statements no longer deprecated

thestelz:
Flip-Flop statements are no longer deprecated per ruby bugs 5400.
Remove deprecated note.

See: https://bugs.ruby-lang.org/issues/5400

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -5334,8 +5334,6 @@ Fix by either:
 
 Avoid the use of flip-flops.
 
-NOTE: They are deprecated as of Ruby 2.6.
-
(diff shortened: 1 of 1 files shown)


thestelz:
Another source, if needed: https://blog.saeloun.com/2020/01/06/ruby-2-7-revert-flip-flop-operator-deprecation.html

pirj:
A PR is welcome.

pirj (approved):
Thanks!

koic:
Good catch!

PR #879, opened 2021-07-25
Fix typo in "Unused Variable Prefix" section

supersam654:
I couldn't figure out what word was supposed to go after "and" so I got rid of it.


koic:
Thanks!

PR #881, opened 2021-08-10
Update README.adoc

alec-c4:
fixed typo


bbatsov:
That's in Bulgarian, so there's no typo. 😆

PR #883, opened 2021-09-08
Fix single-line method rules order

ojab:
`No Single-line Methods` & `Endless Methods` refer one another as
`next` & `previous`, but right now there is `Top-Level Methods`
between them.


koic:
Good catch! Thanks!

PR #888, opened 2021-10-09
move 'if x == 0' to bad examples

ElenaMikhaylova:
Predicate Methods

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -5471,6 +5471,9 @@ end
 if x == nil
 end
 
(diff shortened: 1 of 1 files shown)


pirj:
I'll close this. Please don't hesitate to open a PR that would clarify the existing rule. Having another pair of eyes on those rules to identify ambiguous wording and sources of confusion is very helpful. Thank you for taking the courage to contribute to the guide!

PR #890, opened 2021-11-22
Add recommendations for simple File IO

leoarnold:
In his 2016 talk ["Little Snippets"](https://youtu.be/mC9TyVeER_8?t=560) Xavier Noria (@fxn) points out that Ruby 1.8 had the idiom

```ruby
File.open(filename, 'w') do |f|
  f.write(content)
end
```

whereas starting with Ruby 1.9 this can be replaced with

```ruby
File.write(filename, content)
```

We add this suggestion for simple truncating file writes in text
and binary mode, as well as the corresponding scenaro for reading files.

The suggestions were also implemented for RuboCop:
- `File/Write`: https://github.com/rubocop/rubocop/pull/10260
- `File/Read`: https://github.com/rubocop/rubocop/pull/10261

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2086,6 +2086,72 @@ rescue StandardError => e
 end
 ----
 
+=== Reading from a file [[file-read]]
+
+Use the convenience methods `File.read` or `File.binread` when only reading a file start to finish in a single operation.
+
+[source,ruby]
+----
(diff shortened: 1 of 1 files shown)


bbatsov:
Nice additions! Thanks!

PR #892, opened 2021-12-11
Add new "Hash Literal Values" rule

koic:
## Summary

Ruby 3.1 will be introduced hash shorthand syntax:
https://bugs.ruby-lang.org/issues/17292.

This rule defines hash literal value syntax.

```ruby
# bad
{foo: foo, bar: bar}

# good
{foo:, bar:}
```

## Additional Information

This rule is based on the upgrade from ES5 to ES6 with ESLint.

The ES6 shorthand syntax is chosen by default for ESLint shorthands:

> "always" (default) expects that the shorthand will be used whenever possible.

https://eslint.org/docs/rules/object-shorthand

The rule will also define the new Ruby 3.1 shorthand syntax by default.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4116,6 +4116,19 @@ hash = { :one => 1, :two => 2, :three => 3 }
 hash = { one: 1, two: 2, three: 3 }
 ----
 
+=== Hash Literal Values
+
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #893, opened 2021-12-23
Add new "Block Forwarding" rule

koic:
Ruby 3.1 will be introduced anonymous block forwarding syntax:
https://bugs.ruby-lang.org/issues/11256

```ruby
# bad
def some_method(&block)
  other_method(&block)
end

# good
def some_method(&)
  other_method(&)
end
```

AFAIK, in most cases, block argument is given name similar to `&block` or `&proc`. Their names have no information and `&` will be sufficient for syntactic meaning.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -3101,6 +3101,25 @@ def some_method(...)
 end
 ----
 
+=== Block Forwarding
+
+Use Ruby 3.1's anonymous block forwarding.
+
+In most cases, block argument is given name similar to `&block` or `&proc`. Their names have no information and `&` will be sufficient for syntactic meaning.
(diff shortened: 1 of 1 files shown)


bbatsov:
I think it'd be nice to add the rationale from the PR to the guideline.

koic:
That makes sense! I added the rationale to the guideline.

bbatsov:
Thanks!

marcandre:
We could mention that it adds the information that block is not being captured and only forwarded.

```ruby
# ok
def initialize(&callback)
  # ...
  @callback = callback
end
```

PR #895, opened 2022-04-12
Add new "Object#yield_self" vs "Object#then" rule

ydah:
Follow: https://github.com/rubocop/rubocop/pull/10526

This PR adds "Object#yield_self" vs "Object#then" rule.

Looking at how the method `Object#then` was created, method name `Object#yield_self` states intention, not behavior, so an alias method `Object#then` was created.

Reference: https://bugs.ruby-lang.org/issues/14594

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4394,6 +4394,19 @@ array.reverse.each { ... }
 array.reverse_each { ... }
 ----
 
+=== `Object#yield_self` vs `Object#then` [[object-yield-self-vs-object-then]]
+
+The method `Object#then` is preferred over the method `Object#yield_self`, since method name of `Object#yield_self` states the intention, not the behavior.
+
+[source,ruby]
+----
+# bad
+obj.yield_self { |x| x.do_something }
+
+# good
+obj.then { |x| x.do_something }
+----
+
 == Numbers
 
 === Underscores in Numerics [[underscores-in-numerics]]


README.adoc
```diff
@@ -5727,6 +5727,18 @@ config
...
+
+The method `Object#then` is preferred over the method `Object#yield_self`, since method name of `Object#yield_self` states the intention, not the behavior.
+[source,ruby]
+----
+# bad
+obj.yield_self { |x| x.do_something }
+
+# good
+obj.then { |x| x.do_something }
+----
```
koic, line 5740: It may be better to place it under the following rule.
https://github.com/rubocop/ruby-style-guide#mapfindselectreduceincludesize

ydah: Thank you for your review.
I think it’s better to move it there for sure, so I'll move it there.

ydah:
I updated this PR. Thank you so much!

README.adoc
```diff
@@ -4344,12 +4344,13 @@ def awesome_things(index = nil)
...
 
-=== `map`/`find`/`select`/`reduce`/`include?`/`size` [[map-find-select-reduce-include-size]]
+=== `map`/`find`/`select`/`reduce`/`include?`/`size`/`yield_self` [[map-find-select-reduce-include-size-yield_self]]
 
-Prefer `map` over `collect`, `find` over `detect`, `select` over `find_all`, `reduce` over `inject`, `include?` over `member?` and `size` over `length`.
+Prefer `map` over `collect`, `find` over `detect`, `select` over `find_all`, `reduce` over `inject`, `include?` over `member?`, `size` over `length` and `yield_self` over `then`.
 This is not a hard requirement; if the use of the alias enhances readability, it's ok to use it.
 The rhyming methods are inherited from Smalltalk and are not common in other programming languages.
 The reason the use of `select` is encouraged over `find_all` is that it goes together nicely with `reject` and its name is pretty self-explanatory.
+The reason the use of `then` is encouraged over `yield_self` is that it method name of `yield_self` states the intention, not the behavior.
```
koic, line 4353: I'm sorry about it seems that my explanation was difficult to understand. I think it is better to make separate rules as follows:

```markdown
=== `map`/`find`/`select`/`reduce`/`include?`/`size` [[map-find-select-reduce-include-size]]

Prefer `map` over `collect`, `find` over `detect`, `select` over `find_all`, `reduce` over `inject`, `include?` over `member?` and `size` over `length`.
This is not a hard requirement; if the use of the alias enhances readability, it's ok to use it.
The rhyming methods are inherited from Smalltalk and are not common in other programming languages.
The reason the use of `select` is encouraged over `find_all` is that it goes together nicely with `reject` and its name is pretty self-explanatory.

=== Object#yield_self vs Object#then [[object-yield-self-vs-object-then]]

The method `Object#then` is preferred over the method `Object#yield_self`, since method name of `Object#yield_self` states the intention, not the behavior.
[source,ruby]
----
# bad
obj.yield_self { |x| x.do_something }
# good
obj.then { |x| x.do_something }
----
```

bbatsov:
Thanks!

PR #898, opened 2022-06-27
Add new Atomic File Operations rule

ydah:
Follow: https://github.com/rubocop/rubocop/pull/10696

When doing file operations after confirming the existence check of a file, frequent parallel file operations may cause problems that are difficult to reproduce.
Therefore, it is preferable to use atomic file operations.

```ruby
# bad - race condition with another process may result in an error in `mkdir`
unless Dir.exist?(path)
  FileUtils.mkdir(path)
end

# good - atomic and idempotent creation
FileUtils.mkdir_p(path)

# bad - race condition with another process may result in an error in `remove`
if File.exist?(path)
  FileUtils.remove(path)
end

# good - atomic and idempotent removal
FileUtils.rm_f(path)
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2185,6 +2185,30 @@ File.open('testfile') do |f|
 end
 ----
 
+=== Atomic File Operations [[atomic-file-operations]]
+
+When doing file operations after confirming the existence check of a file, frequent parallel file operations may cause problems that are difficult to reproduce.
+Therefore, it is preferable to use atomic file operations.
+
+[source,ruby]
+----
+# bad - race condition with another process may result in an error in `mkdir`
+unless Dir.exist?(path)
+  FileUtils.mkdir(path)
+end
+
+# good - atomic and idempotent creation
+FileUtils.mkdir_p(path)
+
+# bad - race condition with another process may result in an error in `remove`
+if File.exist?(path)
+  FileUtils.remove(path)
+end
+
+# good - atomic and idempotent removal
+FileUtils.rm_f(path)
+----
+
 === Standard Exceptions [[standard-exceptions]]
 
 Prefer the use of exceptions from the standard library over introducing new exception classes.


README.adoc
```diff
@@ -2185,6 +2185,28 @@ File.open('testfile') do |f|
 end
 ----
 
+Atomic File Operations [[atomic-file-operations]]
```
bbatsov, line 2188: Seems to me you forgot the heading marker here (`===`).

README.adoc
```diff
@@ -2185,6 +2185,28 @@ File.open('testfile') do |f|
...
+
+When file operations after confirming the existence check of a file, frequent parallel file operations may cause problems that are difficult to reproduce.
```
bbatsov, line 2190: when doing

README.adoc
```diff
@@ -2185,6 +2185,28 @@ File.open('testfile') do |f|
...
+Therefore, it is prefer to use atomic file operations.
```
bbatsov, line 2191: it is preferable

README.adoc
ydah: Oops, I added `===`.

README.adoc
ydah: Thank you. I fixed it.

README.adoc
ydah: Thank you. I fixed it.

README.adoc
```diff
@@ -2185,6 +2185,28 @@ File.open('testfile') do |f|
...
+=== Atomic File Operations [[atomic-file-operations]]
+
+When doing file operations after confirming the existence check of a file, frequent parallel file operations may cause problems that are difficult to reproduce.
+Therefore, it is preferable to use atomic file operations.
+
+[source,ruby]
+----
+# bad
+unless File.exist?(path)
+  FileUtils.makedirs(path)
```
pirj, line 2197: https://ruby-doc.org/stdlib-2.7.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-makedirs

> `makedirs`. Alias for: `mkdir_p`.

Why is this a bad example? Would it fail if in between `exists?` and `makedirs` some other process would create the path?

ydah: If the directory existed at the time of `exist?`, but another process erased it immediately afterwards, subsequent processes will be executed without the directory that should have been there.
However, the `makedirs` case is a case that covers a very limited problem, so `Dir.mkdir` may be a more obvious example.

An exception is raised if the directory did not exist at the time of `exist?`, but other process created it before `mkdir` was executed.

README.adoc
```diff
@@ -2185,6 +2185,28 @@ File.open('testfile') do |f|
...
+end
+
+if File.exist?(path)
+  FileUtils.remove(path)
+end
+
+# good
+FileUtils.mkdir_p(path)
+
+FileUtils.rm_rf(path)
```
tdeo, line 2207: I believe the safe alternative that should be suggested for `FileUtils.remove(path) if File.exist?(path)` should rather be `FileUtils.rm_f(path)` as it doesn't have the same behavior than `FileUtils.rm_rf` on directories

ydah: Indeed it is. I changed the example. Thank you so much.

README.adoc
```diff
@@ -2185,6 +2185,28 @@ File.open('testfile') do |f|
...
+[source,ruby]
+----
+# bad
+unless Dir.exist?(path)
+  FileUtils.mkdir(path)
+end
+
+if File.exist?(path)
+  FileUtils.remove(path)
+end
```
pirj, line 2202: It may be understood that those two snippets belong to a single example, while those are two separate ones and put corresponding good examples closer to bad ones.

WDYT of:
```ruby
# bad - race condition with another process may result in an error in `mkdir`
unless Dir.exist?(path)
  FileUtils.mkdir(path)
end

# good - atomic and idempotent creation
FileUtils.mkdir_p(path)

# bad - race condition with another process may result in an error in `remove`
if File.exist?(path)
  FileUtils.remove(path)
end

# good - atomic and idempotent removal
FileUtils.rm_f(path)
```

The above addresses two problems:
1. Confusion with what the example is
2. Order of good and bad

Leaving the order (bad, bad, good, good vs bad, good, bad, good) up to you. I found [one example](https://rubystyle.guide/#other-arg) of the second kind of ordering, but pretty sure one could find examples of bad, bad, good, good, too. So it's subjective and I leave it to your judgement how you want to do it.

pirj (approved):
Thank you!

README.adoc
tdeo: I indeed like the explanation of what concurrency issue can arise, maybe even specify the errors that could happen: `already exists` for `mkdir` and `does not exist` for `remove`?

pirj (approved):

bbatsov (approved):

PR #906, opened 2022-10-18
Add new "Operator Method Call" rule

koic:
Avoid dot where not required for operator method calls.

```ruby
# bad
num.+ 42

# good
num + 42
```

This rule corresponds to https://github.com/rubocop/rubocop/pull/11080.


bbatsov:
Thanks!

PR #907, opened 2022-10-24
Change "Consisten Classes" example to be consistent with style guide

jonaslanglotz:
The example for the "Consistent Classes" section previously used a bad single-line declaration for the inner class, called out two sections below in "Single-line Classes". This PR fixes this inconsistency and applies the recommended two line format.


bbatsov:
Thanks!

Issue #889, opened 2021-10-14
Add guidance for safe navigation operator (`&.`)


andyw8:
Reading https://thoughtbot.com/blog/ruby-safe-navigation reminded me that over-use of `&.` is common. I think it would be helpful to add a guideline such as:

> Avoid long chains of  `&.`. Replace with `.` if they introduce unnecessary conditional logic. Consider other approaches such as delegation.

This could also be added as a rule in RuboCop with a maximum number of chained calls (e.g. 2 or 3).

bbatsov:
Yeah, I agree. We should definitely add some guidelines about `&.`.

Gauravpurohit409:
Created PR for it please review it https://github.com/rubocop/ruby-style-guide/pull/912

pirj:
Fixed in #912

PR #912, opened 2022-12-25
Add guideline for using safe operator.

Gauravpurohit409:
[#889](https://github.com/rubocop/ruby-style-guide/issues/889) - Add guidance for safe navigation operator (&.) 
Based on the 889 issue added guidelines for using safe navigation operators using some resources from the thought bot.

The idea here is to not use a safe navigation operator in unnecessary chaining which may lead to confusion to the reader regarding the existence of the object instead use delegation.

NOTE: This is my first pr so if there is any suggestion please let me know thanks in advance :).

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -325,29 +325,32 @@ foo&.bar
 
 === Safe navigation
 
-Avoid long chains of `&.` replace with `.`, if safe navigator operator introduce unnecessary conditional logic.
-Consider other approaches such as delegation.
+Avoid chaining of `&.`. Replace with `.` and an explicit check.
+E.g. if users are guaranteed to have an address and addresses are guaranteed to have a zip code:
 
-If users are guaranteed to have an address and addresses are guaranteed to have a zip code:
-
-[source, thoughtbot]
+[source, ruby]
 ----
 # bad
 user&.address&.zip
 
 #good 
 user && user.address.zip
+----
+
+If such a change introduces excessive conditional logic, consider other approaches, such as delegation:
+[source, ruby]
+----
+#bad
+user && user.address && user.address.zip
 
 #good
 class User
   def zip
-    address.zip
+    address&.zip
   end
 end
-
 user&.zip
 ----
-
 === Spaces and Braces [[spaces-braces]]
 
 No spaces after `(`, `[` or before `]`, `)`.
--- a/README.adoc
+++ b/README.adoc
@@ -325,7 +325,7 @@ foo&.bar
 
 === Safe navigation
 
-Avoid long chains of `&.`. Replace with `.` if they introduce unnecessary conditional logic. 
+Avoid long chains of `&.`. Replace with `.`. if safe navigator operator introduce unnecessary conditional logic.
 Consider other approaches such as delegation.
 
 For condition where users are guaranteed to have an address and addresses are guaranteed to have a zip code
--- a/README.adoc
+++ b/README.adoc
@@ -323,6 +323,29 @@ foo&. bar
 foo&.bar
 ----
 
+=== Safe navigation
+
+Avoid long chains of `&.`. Replace with `.` if they introduce unnecessary conditional logic. 
+Consider other approaches such as delegation.
+
+For condition where users are guaranteed to have an address and addresses are guaranteed to have a zip code
+
+[source, thoughtbot(https://thoughtbot.com/blog/ruby-safe-navigation#conditional-equivalence)]
+# bad
+user&.address&.zip
+
+#good 
+user && user.address.zip
+
+#good
+class User
+  def zip
+    address.zip
+  end
+end
+
+user&.zip
+
 === Spaces and Braces [[spaces-braces]]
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -323,6 +323,29 @@ foo&. bar
 foo&.bar
 ----
 
+=== Safe navigation
+
+Avoid long chains of `&.`. Replace with `.` if they introduce unnecessary conditional logic.
```
pirj, line 328: "long" sounds vague. How do you feel about removing it altogether?

pirj, line 328: What does it mean "if they introduce unnecessary conditional logic"?

README.adoc
```diff
@@ -323,6 +323,29 @@ foo&. bar
...
 ----
 
+=== Safe navigation
+
+Avoid long chains of `&.`. Replace with `.` if they introduce unnecessary conditional logic. 
+Consider other approaches such as delegation.
+
+For condition where users are guaranteed to have an address and addresses are guaranteed to have a zip code
+
+[source, thoughtbot(https://thoughtbot.com/blog/ruby-safe-navigation#conditional-equivalence)]
```
pirj, line 333: This breaks rendering
image

README.adoc
```diff
@@ -323,6 +323,29 @@ foo&. bar
 foo&.bar
 ----
 
+=== Safe navigation
+
+Avoid long chains of `&.`. Replace with `.` if they introduce unnecessary conditional logic. 
+Consider other approaches such as delegation.
+
+For condition where users are guaranteed to have an address and addresses are guaranteed to have a zip code
```
pirj, line 331:
```suggestion
If users are guaranteed to have an address and addresses are guaranteed to have a zip code:
```

README.adoc
```diff
@@ -323,6 +323,29 @@ foo&. bar
 foo&.bar
 ----
 
+=== Safe navigation
```
pirj, line 326: I'd argue regarding the section, Source code layout.
Maybe it would be a better fit for [Misc](https://rubystyle.guide/#misc)?

README.adoc
Gauravpurohit409: yes sure what alternative would you suggest after removing it?

Gauravpurohit409: Updated the sentence, I mean to say if introducing the safe operator adds unnecessary conditional logic we should use other methods such as delegation.

README.adoc
```diff
@@ -323,6 +323,31 @@ foo&. bar
...
+
+Avoid long chains of `&.` replace with `.`, if safe navigator operator introduce unnecessary conditional logic.
+Consider other approaches such as delegation.
+
+If users are guaranteed to have an address and addresses are guaranteed to have a zip code:
```
pirj, line 331:
```suggestion
Avoid chaining of `&.`. Replace with `.` and an explicit check.
E.g. if users are guaranteed to have an address and addresses are guaranteed to have a zip code:
```

README.adoc
```diff
@@ -323,6 +323,31 @@ foo&. bar
...
+user && user.address.zip
+
+#good
+class User
+  def zip
+    address.zip
+  end
+end
+
+user&.zip
```
pirj, line 348:
```suggestion
----

If such a change introduces excessive conditional logic, consider other approaches, such as delegation:
[source]
----
#bad
user && user.address && user.address.zip

#good
class User
  def zip
    address&.zip
  end
end
user&.zip
```

README.adoc
```diff
@@ -323,6 +323,31 @@ foo&. bar
...
 ----
 
+=== Safe navigation
+
+Avoid long chains of `&.` replace with `.`, if safe navigator operator introduce unnecessary conditional logic.
+Consider other approaches such as delegation.
+
+If users are guaranteed to have an address and addresses are guaranteed to have a zip code:
+
+[source, thoughtbot]
```
pirj, line 333:
```suggestion
[source,ruby]
```
According to the [AsciiDoc documentation](https://docs.asciidoctor.org/asciidoc/latest/verbatim/source-blocks/), the second argument is the language of the source.

No wonder "thoughtbot" here breaks code highlighting:
image

README.adoc
Gauravpurohit409: Can you elaborate more regarding this comment?

pirj: If I'm not mistaken, presently this new guideline resides in the "Source code layout" section. This section is mostly all about whitespace, formatting, those things that don't change code semantics.

The new guideline suggests code changes that may, depending on constraints, change its behaviour.

Basing on this observation, I propose moving this new guideline to a different section. I don't have a better section in mind, so I suggested "Misc".

pirj (approved):
Thank you!

andyw8 (approved):

pirj:
Thank you, @Gauravpurohit409 !

README.adoc
Gauravpurohit409: got it

PR #923, opened 2023-10-28
Fix typo in the Accessor/Mutator Method Names section

ydakuka:


koic:
The example looks correct.

ydakuka:
OK, I thought the "bad" example should be converted to the "good" one.

PR #926, opened 2023-10-29
Fix method order in the Mixin Grouping section

ydakuka:
See https://github.com/rubocop/rubocop/issues/12321#issuecomment-1784010472

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -3289,8 +3289,8 @@ end
 # good
 class Person
   # multiple mixins go in separate statements
-  include Foo
   include Bar
+  include Foo
 end
 ----
 


pirj (changes requested):
The rubocop repo comment says we should not change the order to avoid behaviour change.

Here, we also shouldn’t create an impression that we insist on alphabetical or any ordering.
And the Bar, Foo order feels extremely unnatural and no doubt steals focus from the guideline in question.

I don’t think it makes sense to introduce a rule to sort includes. It would be a nightmare to have all exceptions marked with rubocop:disable.

ydakuka:
Ok, should the PR (https://github.com/rubocop/ruby-style-guide/pull/924) be reverted?

pirj:
No. Kwargs ordering had a clear inconsistency ‘bar’ was a required kwarg as specified in the “bad” example. It should have remained such in the ‘good’ example. This has nothing to do with alphabetic ordering of kwargs.

In any case, kwargs ordering change is cosmetic, while module inclusion order change can cause behaviour change.
Having a alphabetic order is in some cases impossible, and would force users to face hard to track bugs, and in the end inline disable the cop for certain order-sensitive includes.

I’ll close this unless the community has a strong preference for alphabetic order

ydakuka:
Got it, thanks for the explanation.

PR #928, opened 2023-11-03
Fix code in Spaces and Braces

ydakuka:
https://docs.rubocop.org/rubocop/cops_layout.html#layoutspaceinsidehashliteralbraces

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -333,17 +333,17 @@ E.g. if users are guaranteed to have an address and addresses are guaranteed to 
 # bad
 user&.address&.zip
 
-#good 
+# good
 user && user.address.zip
 ----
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -376,11 +376,11 @@ Whichever one you pick - apply it consistently.
...
 [source,ruby]
 ----
+# bad - no space after { and before }
+{one: 1, two: 2}
+
 # good - space after { and before }
 { one: 1, two: 2 }
-
-# good - no space after { and before }
-{one: 1, two: 2}
```
koic, line 383: Both are good cases in this style guide, so this is not typo.

ydakuka: Got it. I've reverted this changes.

pirj:
Thank you for another contribution, @ydakuka !

PR #930, opened 2023-11-04
Add cosmetic changes

ydakuka:


README.adoc
```diff
@@ -214,7 +214,7 @@ Use Unix-style line endings.footnote:[*BSD/Solaris/Linux/macOS users are covered
 ====
 If you're using Git you might want to add the following configuration setting to protect your project from Windows line endings creeping in:
 
-[source,bash]
+[source,shell]
```
ydakuka, line 217: For consistency. I know this is the valid source. https://docs.asciidoctor.org/asciidoc/latest/verbatim/source-highlighter/#shell-vs-console

README.adoc
```diff
@@ -5502,10 +5503,10 @@ u1 = UDPSocket.new
 u1.bind('127.0.0.1', 4913)
 u2 = UDPSocket.new
 u2.connect('127.0.0.1', 4913)
-# Won't send a message to the receiver obj.
-# Instead it will send a message via UDP socket.
+
+# bad - Won't send a message to the receiver obj. Instead it will send a message via UDP socket.
```
koic, line 5507:
```suggestion
# bad - Won't send a message to the receiver object. Instead it will send a message via UDP socket.
```

README.adoc
```diff
@@ -5502,10 +5503,10 @@ u1 = UDPSocket.new
...
 u2.send :sleep, 0
-# Will actually send a message to the receiver obj.
+# good - Will actually send a message to the receiver obj.
```
koic, line 5509:
```suggestion
# good - Will actually send a message to the receiver object.
```

ydakuka: fixed

bbatsov:
Thanks!

README.adoc
pirj: What is worth noticing in the link:
> [for `console`, ] the syntax highlighter will parse the prompt (e.g., $) at the start of each line, then handle the remaining text using the shell language

Should we use `console` here?

koic: Yeah, `console` looks better here.

PR #932, opened 2023-11-15
Fix code in the "Safe navigation" section

ydakuka:
P.S. I think the issue (https://github.com/rubocop/ruby-style-guide/issues/889) should be closed too.


pirj:
The example is correct in correspondence to this note:
> addresses are guaranteed to have a zip code

PR #934, opened 2023-11-15
Merge duplicated rules

ydakuka:
1. the first rule

https://rubystyle.guide/#no-dsl-decorating

https://rubystyle.guide/#declarative-methods-that-have-keyword-status-in-ruby

https://rubystyle.guide/#methods-that-are-part-of-an-internal-dsl

~~2. the second rule~~

~~https://rubystyle.guide/#keyword-arguments-vs-option-hashes~~

~~https://rubystyle.guide/#boolean-keyword-arguments~~

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -814,14 +814,18 @@ user.set(name: 'John', age: 45, permissions: { read: true })
 
 === DSL Method Calls [[no-dsl-decorating]]
 
-Omit both the outer braces and parentheses for methods that are part of an internal DSL.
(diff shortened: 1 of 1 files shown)


bbatsov:
I agree we can merge the 1), but the rules you've listed under 2) seem different. What makes you believe they are the same?

ydakuka:
> but the rules you've listed under 2) seem different. What makes you believe they are the same?

Yes, you're right, I made the mistake here. I've reverted it.

PR #935, opened 2023-11-22
Fix inaccuracy in the Keyword Arguments Order

ydakuka:
Optional arguments are not the same as optional keyword arguments. Since we're talking about keyword arguments here, it's better to clarify this point.


bbatsov:
Thanks!

PR #936, opened 2023-11-22
Fix conflict between `Hash#values_at` and "Parallel Assignment"

ydakuka:


README.adoc
```diff
@@ -2297,7 +2297,7 @@ Prefer the use of exceptions from the standard library over introducing new exce
 === Parallel Assignment [[parallel-assignment]]
 
 Avoid the use of parallel assignment for defining variables.
-Parallel assignment is allowed when it is the return of a method call, used with the splat operator, or when used to swap variable assignment.
+Parallel assignment is allowed when it is the return of a method call, used with the splat operator, or when used to swap variable assignment, or `Hash#values_at` and `Hash#fetch_values`.
```
ydakuka, line 2300: These built-in methods are very convenient, but there are many such methods.

bbatsov: Those are method calls, so it's already covered in this sentence.

README.adoc
```diff
@@ -4404,18 +4404,25 @@ batman.fetch(:powers, obtain_batman_powers) # obtain_batman_powers is an expensi
 batman.fetch(:powers) { obtain_batman_powers }
 ----
 
-=== `Hash#values_at` [[hash-values-at]]
+=== `Hash#values_at` [[hash-values-at-and-fetch-values]]
```
bbatsov, line 4407: You should check if something in RuboCop doesn't refer to this anchor before changing it.

ydakuka: Sure, there are no links to this anchor.

PR #938, opened 2023-12-15
Add `Enumerable#lazy` guideline to "Collections" section

ydakuka:


README.adoc
```diff
@@ -4950,6 +4950,28 @@ LOREM
 "when an unknown printer took a galley of type and scrambled it to make a type specimen book."
 ----
 
+== Enumerable
```
bbatsov, line 4953: Might be good to check if something else can't be moved under this section as well, as it's kind of weird to have a section with a single entry in it.

README.adoc
```diff
@@ -4950,6 +4950,28 @@ LOREM
...
+
+=== lazy [[lazy]]
```
bbatsov, line 4955: I'd use a more descriptive heading here - something like `Leverage lazy to optimize memory usage` or whatever.

README.adoc
```diff
@@ -4950,6 +4950,28 @@ LOREM
...
+
+Use `Enumerable#lazy` in combination with methods like `.first` or `.take`, especially for operations on large collections or datasets, because it avoids the need to process the entire collection.
+
+[source,ruby]
+----
+SIZE = 10
+MAX = 10_000_000
+
+# bad
```
bbatsov, line 4964: You might also want to mention that this will instantiate a huge collection first, so it's clearer what the problem and value proposition is.

ydakuka:
Done. Removed the standalone "Enumerable" section and moved the lazy tip into the "Collections" section — right after collection querying and before "Numbers". It now sits alongside other Enumerable-related tips (flat_map, reverse_each, count vs size, etc.).

PR #939, opened 2024-01-12
Add "Slicing with Ranges" rule

koic:
Follow up https://github.com/rubocop/rubocop/pull/12594#issuecomment-1886375195.

This PR adds "Slicing with Ranges" rule.


bbatsov:
Looks good. Thanks!

PR #941, opened 2024-05-22
Add guideline about omitting super arguments

Earlopain:
For https://github.com/rubocop/rubocop/pull/12427 which is the RuboCop implementation of this. I'll just copy the examples over:

```rb
def some_method(*args, **kwargs)
  super(*args, **kwargs)
end

# good - implicitly passing all arguments
def method(*args, **kwargs)
  super
end

# good - forwarding a subset of the arguments
def method(*args, **kwargs)
  super(*args)
end

# good - calling super with different arguments
def method(*args, **kwargs)
  super("foo", *args, **kwards)
end

# good - forwarding no arguments
def method(*args, **kwargs)
  super()
end
```

There's an interesting interaction with blocks that I wasn't previously aware of: they are just always passed, unless explicitly opted out through `&nil`. Demonstration:

```rb
class A
  def foo(*args, **kwargs)
    if block_given?
      pp yield
    else
      pp "No Block"
    end
  end
  
  def bar(...)
    if block_given?
      pp yield
    else
      pp "No Block"
    end
  end
end

class B < A
  def foo(*args, **kwargs)
    super(*args, **kwargs)
    super()
  end
  
  def bar(*args, **kwargs, &blk)
    super(*args, **kwargs, &blk)
    super()
  end
end

B.new.foo("foo", c: "c", d: "d") { "FOO" }
B.new.bar("bar", c: "c", d: "d") { "BAR" }
# => Never prints "No Block"
```

I've added a callout for this at the end.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -3216,6 +3216,42 @@ def some_method(&)
 end
 ----
 
+=== Super Forwarding
+
+Use `super` without arguments when the enclosing methods arguments are the same.
+
+When super is called without arguments and parentheses Ruby will automatically forward the arguments taken by the method.
+
+[source,ruby]
+----
+# bad
+def some_method(*args, **kwargs)
+  super(*args, **kwargs)
+end
+
+# good - implicitly passing all arguments
+def method(*args, **kwargs)
+  super
+end
+
+# good - forwarding a subset of the arguments
+def method(*args, **kwargs)
+  super(*args)
+end
+
+# good - calling super with different arguments
+def method(*args, **kwargs)
+  super("foo", *args, **kwards)
+end
+
+# good - forwarding no arguments
+def method(*args, **kwargs)
+  super()
+end
+----
+
+IMPORTANT: Blocks are always passed to the parent method. `&nil` can be passed to `super` if the block should remain local.
+
 === Private Global Methods [[private-global-methods]]
 
 If you really need "global" methods, add them to Kernel and make them private.


README.adoc
```diff
@@ -3216,6 +3216,42 @@ def some_method(&)
 end
 ----
 
+=== Super Forwarding
+
```
pirj, line 3220: I’m struggling to understand how this corresponds to the Arguments forwarding section above.

Both that other section and this new one have blind forwarding examples only. There, they suggest using the … syntax, and it is understandable. But only for the blind forwarding case.
But what about cases where only some of the arguments need to be forwarded? Like eg [this one](https://bugs.ruby-lang.org/issues/16253#note-4) with `def m(foo, …)`? Does it even work?

And moving on to the new addition. This all looks reasonable.
My only concerns are below.

README.adoc
```diff
@@ -3216,6 +3216,42 @@ def some_method(&)
...
+----
+# bad
+def some_method(*args, **kwargs)
+  super(*args, **kwargs)
+end
+
+# good - implicitly passing all arguments
+def method(*args, **kwargs)
+  super
+end
```
pirj, line 3235: If this is blind forwarding, should we use the … syntax to describe parameters in the method signature? I understand that it is intent-revealing to write all of them. Won’t RuboCop with the default config suggest prefixing them with `_`?

README.adoc
```diff
@@ -3216,6 +3216,42 @@ def some_method(&)
...
+
+When super is called without arguments and parentheses Ruby will automatically forward the arguments taken by the method.
+
+[source,ruby]
+----
+# bad
+def some_method(*args, **kwargs)
+  super(*args, **kwargs)
+end
+
```
pirj, line 3231: Do we need an example with a named argument? Why are all examples splats?

README.adoc
```diff
@@ -3216,6 +3216,42 @@ def some_method(&)
...
+  super(*args)
+end
+
+# good - calling super with different arguments
+def method(*args, **kwargs)
+  super("foo", *args, **kwards)
+end
+
+# good - forwarding no arguments
+def method(*args, **kwargs)
```
pirj, line 3248: Again, should this be _args and _kwargs?

README.adoc
```diff
@@ -3216,6 +3216,42 @@ def some_method(&)
...
+end
+
+# good - forwarding a subset of the arguments
+def method(*args, **kwargs)
+  super(*args)
+end
+
+# good - calling super with different arguments
+def method(*args, **kwargs)
+  super("foo", *args, **kwards)
```
pirj, line 3244: I haven’t tried, but would `super(“foo”, …)` syntax work?

README.adoc
```diff
@@ -3216,6 +3216,42 @@ def some_method(&)
...
+# bad
+def some_method(*args, **kwargs)
+  super(*args, **kwargs)
+end
+
+# good - implicitly passing all arguments
+def method(*args, **kwargs)
+  super
+end
+
```
pirj, line 3236: How do we dissect when to describe params in details, and when an unnamed splat (`*`) and `…` is fine?

README.adoc
```diff
@@ -3216,6 +3216,42 @@ def some_method(&)
...
+  super("foo", *args, **kwards)
+end
+
+# good - forwarding no arguments
+def method(*args, **kwargs)
+  super()
+end
+----
+
+IMPORTANT: Blocks are always passed to the parent method. `&nil` can be passed to `super` if the block should remain local.
```
pirj, line 3253: What if there are other parameters? What is the recommended way to pass all arguments to super, but not the block? The … won’t work, will it?

pirj (commented):
Thanks! This looks like a nice addition to the guide. I have some questions, though.

The question to when declare parameter names in the method signature or to use the … syntax or an anon splat is not directly related, and would better be a separate guideline if we happen to end up shaping it.

pirj:
@zverok, what is your take on this?

README.adoc
Earlopain: This does work in Ruby 3.1 (I think). `...` got introduced and one or two versions later this was made valid syntax

README.adoc
Earlopain: >I’m struggling to understand how this corresponds to the Arguments forwarding section above.

It fits thematically at least but I see your point about where I placed it. Perhaps it is better suited to come after ["Using `super` with Arguments" ](https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#super-with-args)?

README.adoc
Earlopain: RuboCop is smart enough to not raise an error here: the arguments are not unused

README.adoc
Earlopain: Not sure why I made them all splats to be honest, there's no deeper meaning behind it. I can add a few more examples, though don't really see the reason. There's nothing special about splat/nonsplat and listing all possible variations is unnecessary imo.,  would just consistently use normal arguments instead. Wdyt?

README.adoc
Earlopain: See above

README.adoc
Earlopain: If you don't want to pass the block you write `&nil`, so in these examples it would be `super(*args, **kwargs, &nil)`. `super(..., &nil)` on the other hand is a syntax error so it doesn't look like you are able to suppress the block with that notation.

Should I make a explicit example instead of just mentioning it here?

README.adoc
pirj: My apologies for not expressing my thoughts clearly.
The place is right, I think this new section fits quite well under the arguments forwarding section, and it relates to it.

My concerns were more related to the existing “arguments forwarding” section, as its only “good example” is something I called “blind forwarding”, ie forwarding of all arguments.
This is not exactly always the case, and the existing section is incomplete and may even be confusing due to lack of coverage of those other cases with its only one good example.

Your addition, in the contrary, is detailed, and we happened to discuss a number of edge cases below, so it’s all good.

I’ll keep this comment open for a potential follow-up to improve the “arguments forwarding” section, but not in the scope of this PR.

PR #947, opened 2024-09-11
Extend safe navigation docs about long `&.` chains

### LINKED ISSUE #946: Limit the number of safe navigation calls in a chain
It is painful to see many `&.` calls in a chain.

If I want to have at most 2 safe navigation calls in a row, for example, then:
```ruby
# bad
foo&.bar&.baz&.zoo

# good
foo&.bar&.baz
```

fatkodima:
Closes #946.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -325,23 +325,24 @@ foo&.bar
 
 === Safe navigation
 
-Avoid chaining of `&.`. Replace with `.` and an explicit check.
+Avoid long chains of `&.`. The longer the chain is, the harder it becomes to track what
+on it could be returning a `nil`. Replace with `.` and an explicit check.
 E.g. if users are guaranteed to have an address and addresses are guaranteed to have a zip code:
 
 [source,ruby]
 ----
 # bad
-user&.address&.zip
+user&.address&.zip&.upcase
 
 # good
-user && user.address.zip
+user && user.address.zip.upcase
 ----
 
 If such a change introduces excessive conditional logic, consider other approaches, such as delegation:
 [source,ruby]
 ----
 # bad
-user && user.address && user.address.zip
+user && user.address && user.address.zip && user.address.zip.upcase
 
 # good
 class User
@@ -349,8 +350,9 @@ class User
     address&.zip
   end
 end
-user&.zip
+user&.zip&.upcase
 ----
+
 === Spaces and Braces [[spaces-braces]]
 
 No spaces after `(`, `[` or before `]`, `)`.


pirj (approved):
I love how you put it.

Considering the [law of Demeter](https://en.m.wikipedia.org/wiki/Law_of_Demeter), long chains are inherently bad. But this principle has downsides, and they are due to forcing to write wrappers (like presenters), or propagate increasing the interface of the top-level receiver class.

It really depends. Can it be that an address doesn’t have a zip? It’s business logic, can vary.

This change seems for the best. Can you try finding confirmation that the suggested style is widespread?
Is there a cop already to facilitate finding this ?

fatkodima:
> Can you try finding confirmation that the suggested style is widespread?

At least in my company we are striving to adhere to this. I will try to investigate other codebases.

> Is there a cop already to facilitate finding this ?

I am currently implementing this cop - https://github.com/rubocop/rubocop/issues/13171

fatkodima:
Found only 1 offense in rails - https://github.com/rails/rails/blob/53ce1964f0bd6491fdc08aebc45f61275100f818/actionpack/lib/action_controller/metal/exceptions.rb#L43, so people definitely stick to it

README.adoc
```diff
@@ -351,6 +351,20 @@ class User
...
 ----
+
+=== Long safe navigation chain [[long-safe-navigation-chain]]
+
+Avoid `&.` chains longer than two.
+
+[source,ruby]
+----
+# bad
+user&.address&.zip
```
koic, line 362: I'm unsure if this example is appropriate. According to https://github.com/rubocop/rubocop/pull/13228, up to two consecutive `&.` are allowed by default. It would be better to explicitly clarify, with an explanation and example code, how many consecutive `&.` are considered "Long safe navigation chain".

pirj:
Rails code might not be representative, I believe they are strict in code quality.

How about other repos from real-world-rails-apps and real-world-ruby in general?

bbatsov:
We already have above the following rule:

> Avoid chaining of `&.`. Replace with `.` and an explicit check. Perhaps it just needs to be expanded and clarified a bit instead of adding a new guideline?

fatkodima:
>  Perhaps it just needs to be expanded and clarified a bit instead of adding a new guideline?

Updated existing section for safe navigation. I am bad at writing docs - concrete suggestions welcome.

bbatsov:
Probably we shouldn't mention a specific number here and just expand a bit the rationale - namely that the longer the chain is, the harder it becomes to track what on it could be returning a `nil`. That's always been my fundamental problem with any safe navigation chains.

fatkodima:
Updated with a suggested rationale.

PR #948, opened 2024-10-01
Add section on merging keyword arguments

dvandersluis:
Adds a guideline for rubocop/rubocop#13252. 

```ruby
# bad
some_method(**opts.merge(foo: true))

# good
some_method(**opts, foo: true)
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -3167,6 +3167,19 @@ def some_method(bar: false)
 end
 ----
 
+=== Merging Keyword Arguments [[merging-keyword-arguments]]
+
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -3167,6 +3167,19 @@ def some_method(bar: false)
 end
 ----
 
+=== Merging Keyword Arguments [[merging-keyword-arguments]]
+
+When passing an existing hash as keyword arguments, add additional arguments directly rather than using `merge`.
```
bbatsov, line 3172: It'd be nice to extend this rationale a bit, as it's not clear why this is preferable, but I'll merge this as is for now.

PR #959, opened 2025-03-25
Clarify that `can` is sometimes a valid verb for predicate names

dvandersluis:
Follows https://github.com/rubocop/rubocop/pull/14032#issuecomment-2751535006


bbatsov:
Might be a good idea to add some "OK" example with `can...something?`.

PR #969, opened 2025-09-21
Add section for else branches after loop exit

Zopolis4:
As per https://github.com/rubocop/rubocop/issues/14042#issuecomment-3256898519.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1953,6 +1953,73 @@ end
 end
 ----
 
+=== `else` branches after loop exit [[else-after-loop-exit]]
+
+There is no need to wrap code inside an `else` when the `if` statement skips the rest of the loop.
+
+[source,ruby]
+----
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1953,6 +1953,33 @@ end
...
+  else
+    do_something_else
+  end
+end
+
+# good
+[0, 1, 2, 3].each do |item|
+  if some_condition
+    do_something
+    next
```
pirj, line 1976: Same for `break` and `return`?

Zopolis4: Good catch.

andyw8:
I think there's some nuance to this.

If `some_condition` is acting like a guard clause, then this structure makes sense.

But if it's checking a business condition, the if/else structure may be a better representation of the underlying domain rule.

PR #973, opened 2026-02-01
Add recommendation to avoid variable reflection methods

andyw8:
## Summary

Adds a new guideline recommending against the use of Ruby's variable reflection methods (`instance_variable_get`, `instance_variable_set`, and related methods) as they break encapsulation.

- Lists all relevant instance and class variable reflection methods
- Includes code examples showing the bad practice vs. proper accessor usage
- Notes acceptable use cases (framework/library code, testing utilities, serialization)

## Context

I'm planning to add a new RuboCop cop to enforce this guideline. My motivation is that I often see AI agents use these methods instead of fixing the design.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -5742,6 +5742,54 @@ u2.send :sleep, 0
 u2.__send__ ...
 ----
 
+=== Avoid Variable Reflection Methods [[avoid-variable-reflection]]
+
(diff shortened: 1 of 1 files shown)


andyw8:
Let's start with a smaller scope: https://github.com/rubocop/ruby-style-guide/pull/978

PR #975, opened 2026-02-15
Add guideline for minimizing negations in conditionals

ydakuka:
Recommend preferring the conditional form (if/unless) that results in fewer negations. When negated terms outnumber positive terms, flip the keyword and invert the condition.

Related: https://github.com/rubocop/rubocop/pull/14824

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1736,6 +1736,32 @@ else
 end
 ----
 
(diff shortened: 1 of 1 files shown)


jonas054:
I agree. This is a good idea. It's generally accepted that human brains are ~not good~ **bad** at processing negations, right?

